PR #10, opened 2024-06-04
"develop" functionality

cronan03:


jupyter_builder/base_extension_app.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+class BaseExtensionApp(JupyterApp, DebugLogFileMixin):
+    version = VERSION
+    flags = flags
+    aliases = aliases
+    name = "lab"
+
+    # Not configurable!
+    core_config = Instance(CoreConfig, allow_none=True)
```
fcollonval, line 121: @cronan03 I dived into the code and you should be able to drop this parameter `core_config` (as well as `should_build`, `should_clean` and `splice_source`) as we are not gonna support deprecated commands `install` and `uninstall`.


Dropping the need for `CoreConfig` should reduce the dependencies on `jupyterlab`.

jupyter_builder/jupyter_app.py
```diff
@@ -0,0 +1,325 @@
+# Copyright (c) Jupyter Development Team.
```
fcollonval, line 1: You can remove this file as `JupyterApp` can be imported from `jupyter_core.application` (it is an acceptable dependency).

jupyter_builder/jupyterlab.py
fcollonval, line 1: You should be able to remove this file too as none of the commands here should be part of the builder.

cronan03: @default("labextensions_path")
    def _default_labextensions_path(self):
        lab = LabApp()
        lab.load_config_file()
        return lab.extra_labextensions_path + lab.labextensions_path
   This is a method in base_extension_app.py which requires LabApp() class. And LabApp() is imported from jupyter_builder/jupyterlab.py  . Is this method not required?

fcollonval: We can simplify that code. Those parameters are actually coming from there:

https://github.com/jupyterlab/jupyterlab_server/blob/ea3eba9162b19f1271ab6f2dfd494d85df3ce37e/jupyterlab_server/config.py#L233-L239

With a default value of:

```py
    from jupyter_core.paths import jupyter_path

    @default("labextensions_path")
    def _default_labextensions_path(self) -> list[str]:
        return jupyter_path("labextensions")
```

You can for now mock the value with the default one above (`jupyter_core.paths`) is an acceptable dependency for now.

fcollonval:
@cronan03 was closing this PR intended?

PR #20, opened 2024-07-16
Include jlpm within this package (Issue #15)

### LINKED ISSUE #15: Include jlpm within this package
We should add `jlpm` within this package rather than jupyterlab.

For now, jlpm define there: https://github.com/jupyterlab/jupyterlab/blob/main/jupyterlab/jlpmapp.py
with the entrypoint defined there: https://github.com/jupyterlab/jupyterlab/blob/e048f27548969c0e4403417ac04bc186f119128f/pyproject.toml#L60

cronan03:
Fixes #15


fcollonval:
To fix the linter, you can change the `.pre-commit-config.yaml` file to exclude yarn.js:

for example

```diff
   - repo: https://github.com/codespell-project/codespell
     rev: "v2.2.6"
     hooks:
       - id: codespell
         args: ["-L", "sur,nd"]
+        exclude: jupyter_builder/yarn.js
```

.pre-commit-config.yaml
```diff
@@ -37,6 +37,9 @@ repos:
     hooks:
       - id: codespell
         args: ["-L", "sur,nd"]
+        exclude: jupyter_builder/yarn.js
+        exclude: builder/tsconfig.tsbuildinfo
+        exclude: end-of-file-fixer
```
fcollonval, line 42: Exclude can only be defined once as a regex - you could remove those two lines:

```suggestion
```

Actually you need to add line 20:

```
- id: end-of-file-fixer
  exclude:  tsconfig.tsbuildinfo
```

fcollonval:
Oki I missed that you added the javascript source files within this PR... Could you create another branch for that? This is a good example of why creating multiple branches to avoid mixing up various subjects and making it harder to fix the CI :wink:

PR #23, opened 2024-07-23
Scaffolding for adding a npm package and setting up the releaser

fcollonval:


package.json
```diff
@@ -0,0 +1,135 @@
+{
+    "name": "@jupyterlab/builder",
```
jtpio, line 2: Looks like we would need to be careful, if this gets released, that it does not conflict with the original package released from https://github.com/jupyterlab/jupyterlab.

PR #26, opened 2024-07-30
Update README.md 2

cronan03:
Add mandatory GSoC report to README as it is set as the source of that report - I (@fcollonval) discovered to late some of the requirements so we cannot change the file as the link is frozen in GSoC evaluation database.


fcollonval (changes requested):
Thanks @cronan03 

I have minor points - mainly typos.

cronan03 (commented):
Thanks @fcollonval , the changes have been addressed

fcollonval (approved):
Thanks @cronan03

PR #29, opened 2024-10-22
Migrated Builder Folder to jupyter-builder

### LINKED ISSUE #16: Include @jupyterlab/builder in this package
We should bring `@jupyterlab/builder` within this package and make it generic.

For now the code lives there: https://github.com/jupyterlab/jupyterlab/tree/main/builder

Darshan808:
Fixes #16 

In this PR, I have
- Migrated the `builder` folder from the JupyterLab repository to the j`upyter-builder` package.
- Modified the script to use `jupyter-builder` from the extension’s node_modules if it is presen.

Looking forward to your suggestions for moving forward with jupyter-builder development.


Darshan808:
@JasonWeill
Also, if it's appropriate, could this PR be labeled as 'hacktoberfest-accepted'? It would count toward my contributions, but no worries if not — the main goal is to move the project forward!

Darshan808:
@fcollonval 
I had a couple of questions for clarification:

1. Previously, we used to extract the version of `@jupyterlab/builder` from both the extension’s `package.json` and the staging `package.json` (core) to check for compatibility. Now, will we be extracting the version of `@jupyterlab-builder/builder` from both and checking for compatibility in the same way?

2. Since we plan to replace `jupyterlab` in the build dependencies of the extension template with `jupyter-builder`, does this mean that in a completely new environment, we will be able to build extensions without needing to install the `jupyterlab` package?  If the answer is yes, I have some more doubts, like how can we check core's compatibility without `jupyterlab`?

fcollonval (changes requested):
Thanks for opening this PR @Darshan808 

My first request will be to reorganize the code to not use a subfolder `builder` but directly the existing `src` folder.

This implies to merge the tsconfig.json and package.json into the root one.

Then you will need to uncomment the following block:

https://github.com/jupyterlab/jupyter-builder/blob/f7df010cf5e1d40cf8d69987623eb64072177071/jupyter_builder/federated_extensions.py#L298-L312

and do any changes to make it works; aka checking that the CI is passing.

Don't hesitate to ping me directly on Gitter for faster replies :wink: 
--> https://matrix.to/#/@fcollonval-5b7f8ee9d73408ce4fa5a522:gitter.im

fcollonval:
Regarding your questions

>    Previously, we used to extract the version of @jupyterlab/builder from both the extension’s package.json and the staging package.json (core) to check for compatibility. Now, will we be extracting the version of @jupyterlab-builder/builder from both and checking for compatibility in the same way?

Indeed, we gonna need to check the version of @jupyterlab-builder is compatible between this package and JupyterLab core and between this package and the extension.

>   Since we plan to replace jupyterlab in the build dependencies of the extension template with jupyter-builder, does this mean that in a completely new environment, we will be able to build extensions without needing to install the jupyterlab package? If the answer is yes, I have some more doubts, like how can we check core's compatibility without jupyterlab?

No JupyterLab will still be needed. The end goal of this package is to ease the builder maintenance code and later to support other applications. But the dependency on the application (aka for now JupyterLab) will still be needed in the extension.

To be more concrete, currently an extension for JupyterLab specifies in package.json configuration like:

```json5
"dependencies" {
  "@jupyterlab/application": "^4.0.0"
},
"devDependencies": {
  "@jupyterlab/builder": "^4.0.0"
},
"jupyterlab": {
  "extension": true
}
```

That will be kept - although an additional package will likely be needed:


```diff
"dependencies" {
  "@jupyterlab/application": "^4.0.0"
},
"devDependencies": {
  "@jupyterlab/builder": "^4.0.0",
+  "@jupyterlab/core-meta": "^4.0.0"
},
"jupyterlab": {
  "extension": true
}
```

However, in pyproject.toml, it will be possible to swap jupyterlab by jupyter-builder reducing the build dependencies stack:

```diff
  [build-system]
- requires = ["hatchling>=1.5.0", "jupyterlab>=4.0.0,<5", "hatch-nodejs-version>=0.3.2"]
+ requires = ["hatchling>=1.5.0", "jupyter-builder>=4.0.0,<5", "hatch-nodejs-version>=0.3.2"]
  build-backend = "hatchling.build"
```

Darshan808:
@fcollonval 
Thanks for the clarification! 
I've made the requested changes. To handle version compatibility, I had to bring additional functions from the core's semver.py into jupyterlab_semver.py. Could you please review it now?

Darshan808:
@fcollonval 
Requesting a re-trigger—this should hopefully be the last one for this PR. Thanks for your understanding!

jupyter_builder/jupyterlab_semver.py
```diff
@@ -1076,7 +1076,7 @@ def replace_stars(comp, loose):
 #  1.2.3 - 3.4 => >=1.2.0 <3.5.0 Any 3.4.x will do
 #  1.2 - 3.4 => >=1.2.0 <3.5.0
 def hyphen_replace(mob):
-    from_, fM, fm, fp, fpr, fb, to, tM, tm, tp, tpr, tb = mob.groups()
```
fcollonval: You should not do that as this will modify the algorithm. Prefer ignoring the linter rule for this line.

.eslintrc.js
fcollonval, line 1: Actually we have more modern eslint configuration in `package.json`. So could remove this file and rather solved the raised errors - it is quite easy to fix them:

```
/home/runner/work/jupyter-builder/jupyter-builder/src/build.ts
Error:   127:9   error  'cssImports' is never reassigned. Use 'const' instead  prefer-const
Error:   258:11  error  'jupyterlab' is never reassigned. Use 'const' instead  prefer-const
Error:   258:29  error  'name' is never reassigned. Use 'const' instead        prefer-const
Error:   266:37  error  'schemaDir' is never reassigned. Use 'const' instead   prefer-const
Error:   266:48  error  'themePath' is never reassigned. Use 'const' instead   prefer-const

/home/runner/work/jupyter-builder/jupyter-builder/src/extensionConfig.ts
Error:    13:20  error  Require statement not part of import statement      @typescript-eslint/no-var-requires
Error:    33:16  error  Require statement not part of import statement      @typescript-eslint/no-var-requires
Error:    36:32  error  Require statement not part of import statement      @typescript-eslint/no-var-requires
Error:    37:7   error  'valid' is never reassigned. Use 'const' instead    prefer-const
Error:    72:20  error  Require statement not part of import statement      @typescript-eslint/no-var-requires
Error:   225:7   error  'plugins' is never reassigned. Use 'const' instead  prefer-const
Error:   289:5   error  Move function declaration to function body root     no-inner-declarations

/home/runner/work/jupyter-builder/jupyter-builder/src/webpack-plugins.ts
Error:   92:20  error  Expected { after 'if' condition  curly
```

The tricky one is about `require` statement. You should replace them with import. For example:

```
const baseConfig = require('./webpack.config.base')
```

:arrow_down: 

```
import * as baseConfig from './webpack.config.base'
```

You will also need to explain to TypeScript that those files are raw text; see https://stackoverflow.com/questions/46503615/import-html-from-typescript-with-webpack/47705264#47705264

fcollonval (changes requested):
Thanks @Darshan808 I have two major requests.

.eslintrc.js
Darshan808: @fcollonval 
I ran into a minor issue when switching from `require` to `import` for `metadata_schema.json`. To resolve it, I moved the file under the `src` directory. Is this the right approach, or should I leave it outside `src` and address the issue by modifying include directory in `tsconfig.json` file instead?

jupyter_builder/jupyterlab_semver.py
Darshan808: Oh, I see how TypeScript influenced my thinking  (thought it like object destructuring ). I should have been more mindful.

src/extensionConfig.ts
```diff
@@ -68,7 +70,7 @@ function generateConfig({
   } else if (typeof data.style === 'string') {
     exposes['./style'] = path.join(packagePath, data.style);
   }
-
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
   const coreData = require(path.join(corePath, 'package.json'));
```
fcollonval, line 74: We could change that as `fs` is already a dependency of this library with something like:

```ts
const coreData = JSON.parse(
  fs.readFileSync(
    path.join(corePath, 'package.json'),
    { encoding: 'utf8' }
  )
);
```

src/extensionConfig.ts
```diff
@@ -30,11 +31,12 @@ function generateConfig({
   devtool = mode === 'development' ? 'source-map' : undefined,
   watchMode = false
 }: IOptions = {}): webpack.Configuration[] {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
   const data = require(path.join(packagePath, 'package.json'));
```
fcollonval, line 35: We could change that as `fs` is already a dependency of this library with something like:

```ts
const coreData = JSON.parse(
  fs.readFileSync(
    path.join(packagePath, 'package.json'),
    { encoding: 'utf8' }
  )
);
```

See https://nodejs.org/api/fs.html#fsreadfilesyncpath-options for the api.

fcollonval (commented):
Thanks @Darshan808

I have two alternative suggestions to avoid skipping linter rule for `require()`. And then we are in a good shape to merge this.

A good test to ad would be the matching of the `@jupyterlab/builder`, would you be up to the task? We could do that in a follow-up PR.

Darshan808:
> A good test to ad would be the matching of the `@jupyterlab/builder`, would you be up to the task? We could do that in a follow-up PR.
@fcollonval 
Yes, I’m up to the task! Could you please clarify what you mean by ‘matching of the `@jupyterlab/builder`? I just want to make sure I understand it fully. Thank you!

fcollonval:
> Yes, I’m up to the task! Could you please clarify what you mean by matching of the @jupyterlab/builder?

The version matching you restored in this PR. Do you see how to do that?

fcollonval:
> Regarding the test you're mentioning, what specific version are we matching the @jupyterlab/builder against? Could you provide some clarification on this?

The test done in the code is a version range matching between the `@jupyterlab/builder` package version shipped within this library at https://github.com/jupyterlab/jupyter-builder/blob/f7df010cf5e1d40cf8d69987623eb64072177071/package.json#L3 and the range provided by the extension at https://github.com/jupyterlab/extension-template/blob/5696924f5b7adf07cf07ea9d3510c66914fbb770/template/package.json.jinja#L68.

So the test should have the following steps:
- Set up a extension from the template
- Modify the range of @jupyterlab/builder to be incompatible with the current version
- Check that the expected error is raised


Existing test are located at https://github.com/jupyterlab/jupyter-builder/blob/main/tests/test_tpl.py
That should help you with the first step. Let me know if you need help for the others steps.

Darshan808:
> That should help you with the first step. Let me know if you need help for the others steps.

Understood! With the major festive season in Nepal, it's a bit challenging to focus fully at the moment, but I’ll make sure to complete it in the next few days.

Darshan808:
#### Added a test to validate version compatibility. 

The test modifies the version range for `@jupyterlab/builder` in the extension then attempts to build the extension.
During this process, we verify that the expected error related to version mismatches is raised.

**Note:** I have currently set the extension's version range to exactly `4.0.0`. This can be adjusted for optimal compatibility.

Please review and suggest any changes or modifications.

Darshan808:
@fcollonval 
I'm getting the expected warning locally for the version matching test and the test passes locally. Not sure why it’s failing on CI, but I’m looking into it. Any insights would be helpful!

Darshan808:
@fcollonval 
Tests are all passing now! Could you review the code and share any feedback?

fcollonval (changes requested):
Thanks  a lot @Darshan808 for the test - and congrats to make it work right away.

I have a minor suggestion. But otherwise we are good to go.

tests/test_tpl.py
```diff
@@ -162,3 +166,45 @@ def test_watch_functionality(tmp_path):
...
+
+    package_json_path = extension_folder / "package.json"
+
+    # Modify the @jupyterlab/builder version to an incompatible range
+    with package_json_path.open("r+", encoding="utf-8") as f:
+        package_data = json.load(f)
+        package_data["devDependencies"]["@jupyterlab/builder"] = "4.0.0"
+        f.seek(0)
+        json.dump(package_data, f, indent=2)
+        f.truncate()
```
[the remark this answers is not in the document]
Darshan808: Thanks for the suggestion! I made the changes, and it looks much better now.

fcollonval (approved):
Thanks a lot for the patience @Darshan808 

Congrats on the first PR for this repo.

Darshan808:
@fcollonval 
Just wanted to say a huge thank you for all the guidance and support throughout this PR. Mentors like you make such a big difference, especially for us newer open-source contributors—it keeps us motivated to keep pushing forward.

Since I’m now part of the JupyterLab triage team, I’m even more excited to keep contributing! What's next for us here, or do we have any plans on when core might start using this package?

jtpio:
Thanks @Darshan808 for the contribution and @fcollonval for the guidance!

> What's next for us here, or do we have any plans on when core might start using this package?

Now that JupyterLab `4.3.0`, there is active work on the next `4.4.0` version.  I haven't caught up with the latest activity on this `jupyter-builder` project, but maybe it would be a good candidate for the `4.4.0` release?
Doing the switch sooner than later would also help avoid having to maintain the same code in two different places, and avoid potential divergences.

Darshan808:
Sounds like a great idea, and I'm definitely on board! I believe I can contribute effectively with a bit of guidance from @fcollonval. Here are some steps i think needs to be done:

- Add jupyter-builder as a dependency in the core
- Update the build scripts to utilize jupyter-builder
- Remove the existing builder from the core

I would appreciate any feedback on this and guidance on the steps I should take to move forward!

Darshan808:
@fcollonval
Should I open an issue here or in the JupyterLab repo to discuss transitioning the core to use this builder? When are we planning to start this process? If you could provide a structure or roadmap, I’d be happy to start working on it.

PR #57, opened 2026-03-15
Use `@jupyterlab/core-meta` package

Darshan808:
### Description
https://www.npmjs.com/package/@jupyterlab/core-meta was published recently.
We now use this package to obtain core's metadata.

As we added a new option `core-package-file`. The builder now fails to build extensions and throws error:
```js
jupyter-builder build .
Building extension in .
error: unknown option '--core-package-file'
```

I think that once this PR is finalized, it would be a good time to say goodbye to the `builder` (npm package) from the JupyterLab core repository and start publishing it from here instead. Otherwise, we won’t be able to test it with `extension-template` and other repositories without publishing the `builder` npm package from this repository.

### DIFF
--- a/jupyter_builder/federated_extensions.py
+++ b/jupyter_builder/federated_extensions.py
@@ -334,9 +334,7 @@ def _ensure_builder(ext_path, core_package_file):
         msg = f"Extensions require a devDependency on @jupyterlab/builder@{dep_version1}, you have a dependency on {dep_version2}"  # noqa: E501
         raise ValueError(msg)
 
-    return osp.join(
-        target, "node_modules", "@jupyterlab", "builder", "lib", "build-labextension.js"
-    )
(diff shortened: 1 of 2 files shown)


jupyter_builder/core_path.py
```diff
@@ -1,25 +1,50 @@
...
+
+    Prefer an installed ``@jupyterlab/core-meta`` package when available.
+    Otherwise, fetch and cache the staging ``package.json`` from GitHub.
     """
 
+    if ext_path is not None:
+        installed_core_meta = _get_installed_core_meta(Path(ext_path).resolve())
+        if installed_core_meta is not None:
+            return installed_core_meta
+
```
Darshan808: I think we should prioritize `version` if given. This'll let advance contributors to build against different core versions regardless of `core.package.json`.

Issue #41, opened 2026-01-21
Renaming package


Darshan808:
Currently, the name of the npm package is `@jupyterlab/builder`
https://github.com/jupyterlab/jupyter-builder/blob/7f1b12b4a0816666243e28f711badd51f956a0db/package.json#L2

which conflicts with the core builder package.
https://github.com/jupyterlab/jupyterlab/blob/dbc512a35ec34456d840c4331298085056896a7c/builder/package.json#L2

In the future, after this builder is ready we want to remove the core's builder package entirely, but to avoid confusion, it might make sense to give this package a distinct name. Some possible alternatives: `@jupyter/extensionn-builder` or `@jupyterlab/extension-builder`.

I believe this would help clarify the purpose of the package and avoid conflicts with the core builder package.

Darshan808:
There is also a npm package https://www.npmjs.com/package/@jupyterlab/extension-builder which was last published 9 years ago.

krassowski:
> Wasn't the idea to move the `@jupyterlab/builder` from core to this repo at some point? So it keeps the same name, but is now developed here instead?

It would be reasonable to move without renaming, as long as it does not cause issues with releases:
- we would need to bump version here to some 4.6.0aX:
   https://github.com/jupyterlab/jupyter-builder/blob/7f1b12b4a0816666243e28f711badd51f956a0db/package.json#L2-L3
- we would need to configure trusted publisher on npm to allow publishing from both repos - from this one for `4.6.x` line and from `jupyterlab` for `4.5.x` line - but since this is not currently possible, we would need to use tokens to publish from this repo I guess:
  > Note: Each package can only have one trusted publisher configured at a time. (https://docs.npmjs.com/trusted-publishers)

krassowski:
Related to https://github.com/jupyterlab/jupyter-builder/issues/49 - we cannot really start publishing this as `4.6.x` right now as this repo is still using `webpack` and the other one is using `rspack`

If we renamed the `npm` package we could use `4.5.x` tag for `webpack` and then `4.6.x` for `rspack` versions. Or we could call them `v1` and `v2`.

PR #60, opened 2026-03-28
Rename Package

Darshan808:
### Description

Rename Package from `@jupyterlab/builder` to `@jupyter/builder` as discussed earlier in #41


krassowski (commented):
Do we need to update other occurrences? I think yes for:
- `.copier-answers.yaml`
- `src/build.ts`

but not sure about others (I have not looked in depth)

Image

PR #69, opened 2026-04-01
Setup Proper Documentation

### LINKED ISSUE #46: Setup proper documentation
### Problem

Since this is going to be a main tool to build extensions we should have a good documentation for it.

### Proposed Solution

- Move some snippets from JupyterLab docs
- Setup spinhx

### Additional context

We could move https://github.com/jupyterlab/jupyter-builder?tab=readme-ov-file#gsoc-2024-anecdote out of the readme to a separate file/branch

Darshan808:
### Fixes #46

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,15 +23,15 @@ Compile the extension JavaScript assets for consumption by a Jupyter app.
 jupyter-builder build [options] <path/to/extension>
 ```
 
-| Flag | Description |
-|---|---|
-| `--development` | Build in development mode (default: `False`) |
-| `--source-map` | Generate source maps (default: `False`) |
-| `--static-url=<url>` | Set the URL for static assets |
-| `--core-version=<version>` | JupyterLab core version to build against |
+| Flag                         | Description                                                            |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -5,104 +5,119 @@
...
 ```
 
-## What does it do?
+| Flag | Description |
+|---|---|
+| `--development` | Build in development mode (default: `False`) |
+| `--source-map` | Generate source maps (default: `False`) |
+| `--static-url=<url>` | Set the URL for static assets |
+| `--core-version=<version>` | JupyterLab core version to build against |
+| `--core-package-file=<path>` | Path to a core application `package.json` (overrides `--core-version`) |
```
Darshan808, line 66: Do we expose `--core-package-file` to users? Or should we remove this flag as we already have `--core-version`?

krassowski: I think it is better to document everything, but maybe we could mark it as (advanced) to highlight that this is not intended as the primary way of configuring things?

krassowski (approved):

PR #72, opened 2026-04-09
Remove `jupyterlab-server` as a dependency

### LINKED ISSUE #48: Should jupyter-builder include `jupyterlab_server.process`?
### Problem

#44 adds a dependency on `jupyterlab_server` because of:

https://github.com/jupyterlab/jupyter-builder/blob/7f1b12b4a0816666243e28f711badd51f956a0db/jupyter_builder/jlpm.py#L9

However, we would like this tool to be as lightweight as possible. `jupyterlab-server` has heavy set of dependencies.

### Proposed Solution

Move [`process`](https://github.com/jupyterlab/jupyterlab_server/blob/f64f554291a09f072e479ff52ae2212084aaac39/jupyterlab_server/process.py#L1) from `jupyterlab-server` to `jupyter-builder`?

### Additional context

This would mean that `jupyterlab-server` would depend on `jupyter-builder`.

Darshan808:

### DIFF
--- a/jupyter_builder/jlpm.py
+++ b/jupyter_builder/jlpm.py
@@ -5,34 +5,30 @@
 
 import os
 import subprocess
-
-# Copyright (c) Jupyter Development Team.
-# Distributed under the terms of the Modified BSD License.
 import sys
-from shutil import which as _which
+from shutil import which
 
 HERE = os.path.dirname(os.path.abspath(__file__))
 YARN_PATH = os.path.join(HERE, "yarn.js")
 
 
-def which(command: str, env: dict[str, str] | None = None) -> str:
-    """Get the full path to a command.
+def _which_node_js(env: dict[str, str] | None = None) -> str:
+    """Get the full path to Node.js executable
 
     Parameters
     ----------
-    command: str
-        The command name or path.
     env: dict, optional
         The environment variables, defaults to `os.environ`.
     """
+    command = "node"
     env = env or os.environ  # type:ignore[assignment]
     path = env.get("PATH") or os.defpath  # type:ignore[union-attr]
-    command_with_path = _which(command, path=path)
+    command_with_path = which(command, path=path)
 
     # Allow nodejs as an alias to node.
     if command == "node" and not command_with_path:
         command = "nodejs"
-        command_with_path = _which("nodejs", path=path)
+        command_with_path = which("nodejs", path=path)
 
     if not command_with_path:
         if command in ["nodejs", "node", "npm"]:
@@ -48,7 +44,7 @@ def which(command: str, env: dict[str, str] | None = None) -> str:
     return os.path.abspath(command_with_path)
 
 
-def execvp(cmd, argv):
+def _execvp_node(argv):
     """Execvp, except on Windows where it uses Popen.
 
     The first argument, by convention, should point to the filename
@@ -57,10 +53,9 @@ def execvp(cmd, argv):
     Python provides execvp on Windows, but its behavior is problematic
     (Python bug#9148).
     """
-    cmd = which(cmd)
+    cmd = _which_node_js()
     if os.name == "nt":
         import signal
-        import sys
 
         p = subprocess.Popen([cmd] + argv[1:])  # noqa S603
         # Don't raise KeyboardInterrupt in the parent process.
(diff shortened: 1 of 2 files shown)


jupyter_builder/jlpm.py
```diff
@@ -1,17 +1,53 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
 """A Jupyter-aware wrapper for the yarn package manager"""
 
 import os
+import subprocess
 
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
```
krassowski: Let's tidy up as we go

```suggestion
```

jupyter_builder/jlpm.py
```diff
@@ -1,17 +1,53 @@
...
+            msg = (
+                "Please install Node.js and npm before continuing installation. "
+                "You may be able to install Node.js from your package manager, "
+                "from conda, or directly from the Node.js website "
+                "(https://nodejs.org)."
+            )
+            raise ValueError(msg)
+        msg = f"The command was not found or was not executable: {command}."
+        raise ValueError(msg)
+    return os.path.abspath(command_with_path)
```
krassowski, line 39: It looks like we could simplify that slightly (and avoid exposing it as public):


```suggestion
def _which_node_js(env: dict[str, str] | None = None) -> str:
    """Get the full path to Node.js executable

    Parameters
    ----------
    env: dict, optional
        The environment variables, defaults to `os.environ`.
    """
    command = "node"
    env = env or os.environ  # type:ignore[assignment]
    path = env.get("PATH") or os.defpath  # type:ignore[union-attr]
    command_with_path = _which(command, path=path)

    # Allow nodejs as an alias to node.
    if not command_with_path:
        command = "nodejs"
        command_with_path = _which("nodejs", path=path)

    if not command_with_path:
        msg = (
            "Please install Node.js and npm before continuing installation. "
            "You may be able to install Node.js from your package manager, "
            "from conda, or directly from the Node.js website "
            "(https://nodejs.org)."
        )
        raise ValueError(msg)
    return os.path.abspath(command_with_path)
```

jupyter_builder/jlpm.py
krassowski:
```suggestion
    cmd = _which_node_js("node")
```

jupyter_builder/jlpm.py
```diff
@@ -1,17 +1,53 @@
...
+
+
 def execvp(cmd, argv):
```
krassowski:
```suggestion
def _execvp_node(argv):
```

jupyter_builder/jlpm.py
krassowski:
```suggestion
    _execvp_node(["node", YARN_PATH, *argv])
```

jupyter_builder/jlpm.py
```diff
@@ -26,7 +62,7 @@ def execvp(cmd, argv):
         import signal
         import sys
```
krassowski: This is also imported at the top

```suggestion
```

jupyter_builder/jlpm.py
```diff
@@ -1,17 +1,53 @@
...
 
 import os
+import subprocess
 
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import sys
-
-from jupyterlab_server.process import subprocess, which
+from shutil import which as _which
```
krassowski: After renaming `which` to `_which_node_js` this import can be simplified too to avoid renaming.

krassowski (commented):
Looks good, but we could take the opportunity to clean it up and reduce API surface for the future.

jupyter_builder/jlpm.py
```diff
@@ -1,18 +1,50 @@
...
+    env: dict, optional
+        The environment variables, defaults to `os.environ`.
+    """
+    command = "node"
+    env = env or os.environ  # type:ignore[assignment]
+    path = env.get("PATH") or os.defpath  # type:ignore[union-attr]
+    command_with_path = which(command, path=path)
+
+    # Allow nodejs as an alias to node.
+    if command == "node" and not command_with_path:
```
krassowski:
```suggestion
    if not command_with_path:
```

jupyter_builder/jlpm.py
```diff
@@ -1,18 +1,50 @@
...
+        command = "nodejs"
+        command_with_path = which("nodejs", path=path)
+
+    if not command_with_path:
+        if command in ["nodejs", "node", "npm"]:
```
krassowski: Can we simplify this further?

krassowski (approved):

PR #73, opened 2026-04-09
Restore `core-path` flag as mark it as depricated.

Darshan808:
### Description
As discussed here https://github.com/jupyterlab/jupyterlab/pull/18723#discussion_r3058999024, some downstream application might be using `core-path`. So marking as depricated in python side.


krassowski (approved):
Thank you!

PR #74, opened 2026-04-10
Add `jupyter-core` as a dependency

Darshan808:
### Description
Identified in the failing action: https://github.com/jupyterlab/jupyterlab/actions/runs/24234580389/job/70753875502?pr=18723
with error:
```
File "/private/var/folders/tb/y368xp_x10s3ty1b_mtl5mxr0000gn/T/pip-build-env-r_9s_1o8/overlay/lib/python3.14/site-packages/jupyter_builder/main.py", line 6, in <module>
  from jupyter_core.application import JupyterApp
ModuleNotFoundError: No module named 'jupyter_core'
```
`jupyter-builder` uses `jupyter_core` but does not declare it as dependency.
https://github.com/jupyterlab/jupyter-builder/blob/c6754b14e38cba51141e5ae4ec918b5bfeb1cc02/jupyter_builder/main.py#L6

But the CI should've failed on #72 
- [ ] Add a test for it.
- [ ] See if we could just import Application from traitlets instead

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Maintenance and upkeep improvements
 
-- Restore `core-path` flag as mark it as depricated. [#73](https://github.com/jupyterlab/jupyter-builder/pull/73) ([@Darshan808](https://github.com/Darshan808), [@krassowski](https://github.com/krassowski))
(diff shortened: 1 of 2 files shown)


krassowski (approved):
Let's add it for now but also open a new issue to consider switching to just using `traitlets.Application` instead (and describe why we need/don't need `jupyter-core` there)

PR #75, opened 2026-04-10
Update project URLs in pyproject.toml

krassowski:
We updated them in `package.json` already but forgot about `pyproject.toml`


Darshan808 (approved):
Thanks!
Failing lint is addressed in #74

PR #78, opened 2026-04-10
Remove `jupyterlab` from test dependencies and drop `jupyter_sever` usage.

### LINKED ISSUE #77: Remove usage of `jupyter_server` as well.
### Description
There is a usage of `jupyter_server` in the codebase.
https://github.com/jupyterlab/jupyter-builder/blob/25971babdcece629f750dff38a023c4049a8518b/jupyter_builder/federated_extensions.py#L26

which is just:
```python

class ArgumentConflict(ValueError):
    pass
```

Darshan808:
### Description

- Remove `jupyterlab` from test dependencies.
- Modify tests to not use `jupyterlab` (as of now the build script has `jupyter labextension`. So don't use build command for now.
- Remove `jupyter_server` usage and declare the imported error locally

### DIFF
--- a/jupyter_builder/federated_extensions.py
+++ b/jupyter_builder/federated_extensions.py
@@ -23,7 +23,6 @@ from os.path import join as pjoin
 
 from jupyter_core.paths import ENV_JUPYTER_PATH, SYSTEM_JUPYTER_PATH, jupyter_data_dir
 from jupyter_core.utils import ensure_dir_exists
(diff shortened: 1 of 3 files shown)


Darshan808:
Now the tests are failing!
We earlier had `jupyterlab` as a test dependency which pulled in everything and we never knew any issues.

krassowski (approved):
Interestingly, `ArgumentConflict` is not used no caught in jupyter-server so I don't know why it is even defined there.

PR #79, opened 2026-04-10
Add Licenses

Darshan808:
### Description
`jupyterlab` has **Semver File License** at the end of its [License file](https://github.com/jupyterlab/jupyterlab/blob/main/LICENSE). Since semver file now also live here as [jupyterlab_semver.py](https://github.com/jupyterlab/jupyter-builder/blob/main/jupyter_builder/jupyterlab_semver.py), moving the license here as well.

I'm not sure on how we can address licenses of [yarn.js](https://github.com/jupyterlab/jupyter-builder/blob/main/jupyter_builder/yarn.js). Currently licenses are listed at the end of yarn.js file in this manner:
```
/**
  @license
  The MIT License (MIT)

  Copyright (c) 2014 Blake Embrey (hello@blakeembrey.com)

  Permission is hereby granted ......
  ```
  I've currently added them on `LICENSE` file but we can discuss on the ideal way to handle these.


krassowski (commented):
Can you use `license-files` from PEP 639 https://peps.python.org/pep-0639/#add-license-files-key

See also discussion in https://github.com/jupyterlab/jupyterlab/pull/14046.

So there are two nice things:
- `jupyter-builder` will simplify LICENCE of `jupyterlab` repo
- PEP 639 was approved which I did not notice earlier

krassowski (approved):

PR #83, opened 2026-04-21
Fix missing `tomli` dependency for Python < 3.11

jtpio:
As noticed in https://github.com/jupyter/notebook/pull/7893#issuecomment-4288784413


jtpio:
Python 3.10 is still around for a couple of months, so probably we can have that dependency until its end of life.

krassowski (approved):
Thanks!

Issue #81, opened 2026-04-11
Support `@jupyterlab/builder` fallback or proceed with #71?


Darshan808:
### Description

After the merge and release of https://github.com/jupyterlab/jupyterlab/pull/18723, the extension build flow changes:

- `jlpm run build` → `jupyter-builder build .`  
- `jupyter labextension build` (deprecated) → proxies to `jupyter-builder build`

`jupyter-builder` then looks for the npm package `@jupyter/builder` inside the extension’s `node_modules` to execute the JS build scripts.

However, existing extensions do not yet depend on `@jupyter/builder`, and the transition will take time. This leads to build failures in current workflows.

### Reproduction

In an environment where:
- JupyterLab is installed in editable mode
- Working on the same branch as https://github.com/jupyterlab/jupyterlab/pull/18723

Switch to an extension directory and run:

```bash
jlpm run build                                                 ─╯
(Deprecated) 'jupyter labextension build' is deprecated, use 'jupyter-builder build' instead.

Building extension in .
node:internal/modules/cjs/loader:1424
  throw err;
  ^

Error: Cannot find module '.../node_modules/@jupyter/builder/lib/build-labextension.js'
```

### Current workaround

Manually adding `@jupyter/builder` to `devDependencies` resolves the issue:
```
jlpm add -D @jupyter/builder
```
### Problem

Once JupyterLab is released with `jupyter-builder`, builds will rely on `@jupyter/builder`, but most extensions will not have it installed yet. This can break extension builds during the transition period.

### Questions / Options

- Should we fallback to `@jupyterlab/builder` if `@jupyter/builder` is not found?
- Or should we proceed with the approach in #71 (packaging the TypeScript build scripts within the Python package)?

Would like guidance on the preferred transition strategy.

krassowski:
> However, existing extensions do not yet depend on @jupyter/builder,

I assume you mean third-party extensions like say `jupyterlab-favourites`? So the CI for them would break.

> Should we fallback to @jupyterlab/builder if @jupyter/builder is not found?

I think this would cause confusion in the future.

> Or should we proceed with the approach in https://github.com/jupyterlab/jupyter-builder/issues/71 (packaging the TypeScript build scripts within the Python package)?

I think this would be the cleanest way forward. This is similar to how we ship static JS assets for JupyterLab core and how each extension ships webpack/rspack-compiled federated extensions JS assets.

[...]

Issue #71, opened 2026-04-02
Bundle and ship TypeScript build scripts within the `jupyter-builder` Python package


Darshan808:
> Is it really expected? Or should jupyter-builder ship with @jupyter/builder assets already?

The current flow is to invoke the Python script via `jupyter labextension build .`, which internally calls the build script from:
```
extension/node_modules/@jupyterlab/builder
```

The new flow changes this to invoking `jupyter-builder build .`, which calls the build script from:
```
extension/node_modules/@jupyter/builder
```

I think it is possible to ship the Python package with the TypeScript build scripts bundled, and invoke them directly. This would remove the need for extensions to depend on an npm package and look into their own `node_modules` for the build script.

This is worth discussing further, since it impacts the overall build flow and I think it has its own pros and cons.

_Originally posted by @Darshan808 in https://github.com/jupyterlab/extension-template/issues/133#issuecomment-4175992353_

PR #87, opened 2026-04-22
Handle both `@jupyter/builder` and `@jupyterlab/builder`

jtpio:
As noticed in https://github.com/jupyter/notebook/pull/7893#discussion_r3122020901

### DIFF
--- a/jupyter_builder/federated_extensions.py
+++ b/jupyter_builder/federated_extensions.py
@@ -301,18 +301,31 @@ def watch_labextension(  # noqa: PLR0913
 # ------------------------------------------------------------------------------
 
 
+# Marker packages an extension may declare to identify its builder, in order
+# of preference.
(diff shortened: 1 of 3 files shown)


krassowski:
I'm not sure if we want to support both https://github.com/jupyterlab/jupyter-builder/issues/81 and #71 are related.

jtpio:
> both #81 and #71 are related.

Right thanks for linking to those issues.

I opened this draft PR to start thinking more about this and check what it would look like in practice. But yeah one of those approaches may indeed be preferred.

krassowski (approved):
Thank you!

Darshan808 (approved):
Thanks

Darshan808:
Released in `v0.0.8`.

jtpio:
Thanks!

PR #89, opened 2026-04-24
Bundle static `js` files and `node_modules` in the python package

Darshan808:
### Description

Added scripts to bundle the `lib/` directory together with `node_modules/` into the Python package.

While this works, the packaged `node_modules` includes platform-specific binaries (`binding-darwin-arm64/rspack.darwin-arm64.node` on my macOS).

This means platform-specific wheels per architecture will be required.


Darshan808:
Not sure if we want to move forward with this approach, but it’s one option that works reliably.

PR #91, opened 2026-05-01
Add downstream build tests

### LINKED ISSUE #88: Downstream tests for JupyterLab and Notebook
Should we build here too? My thinking is mostly about enabling work on https://github.com/jupyterlab/jupyter-builder/issues/70 without fear of breaking downstreams.

Darshan808:
### Description
Add a new workflow to build `jupyterlab` and `notebook` in CI.

### DIFF
--- a/.github/workflows/downstream-tests.yml
+++ b/.github/workflows/downstream-tests.yml
@@ -22,7 +22,9 @@ jobs:
       uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
 
     - name: Install Builder
-      run: python -m pip install .
+      run: |
+        pip install hatchling>=1.21.1
+        python -m pip install .
 
     - name: Install dependencies
       run: jlpm install
@@ -47,7 +49,10 @@ jobs:
           '.resolutions["@jupyter/builder"] = $t' \
           downstream/package.json > tmp.json && mv tmp.json downstream/package.json
 
-    - name: Install JS deps
+        echo "--- Patched resolution ---"
+        jq '.resolutions["@jupyter/builder"]' downstream/package.json
+
+    - name: Install deps
       working-directory: downstream
       run: jlpm install
 
--- a/.github/workflows/downstream-tests.yml
+++ b/.github/workflows/downstream-tests.yml
@@ -59,3 +59,16 @@ jobs:
     - name: Build ${{ matrix.name }}
       working-directory: downstream
       run: jlpm build
+
+    - name: Build wheel
+      working-directory: downstream
+      run: |
+        pip install build
+        python -m build --wheel
+
+    - name: Upload wheel
+      uses: actions/upload-artifact@v4
+      with:
+        name: ${{ matrix.name }}-wheel
+        path: downstream/dist/*.whl
+        if-no-files-found: error
--- a/.github/workflows/downstream-tests.yml
+++ b/.github/workflows/downstream-tests.yml
@@ -1,0 +1,56 @@
+name: Downstream Tests
+
+on:
+  pull_request:
+
+jobs:
+  downstream:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - repo: jupyterlab/jupyterlab
+            name: jupyterlab
+          - repo: jupyter/notebook
+            name: notebook
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v6
+
+    - name: Base Setup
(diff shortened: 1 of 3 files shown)


Darshan808:
Downstream tests are passing.
Now just to confirm that it is using locally built `@jupyter/builder`, let me try and err out.

Darshan808:
> /home/runner/work/jupyter-builder/jupyter-builder/downstream/node_modules/@jupyter/builder/lib/build-labextension.js:90
    throw new Error('Test');
    
This confirms it.

.github/workflows/downstream-tests.yml
```diff
@@ -0,0 +1,61 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v6
+
+    - name: Base Setup
+      uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
+
+    - name: Install Builder
+      run: |
+        pip install hatchling>=1.21.1
```
krassowski, line 26: Do we need this line:

```suggestion
```

Given:

https://github.com/jupyterlab/jupyter-builder/blob/a140a88fc8b2d04b648cef9c2d6712713c5a9d04/pyproject.toml#L4-L6

.github/workflows/downstream-tests.yml
```diff
@@ -0,0 +1,61 @@
...
+        echo "--- Patched resolution ---"
+        jq '.resolutions["@jupyter/builder"]' downstream/package.json
+
+    - name: Install deps
+      working-directory: downstream
+      run: jlpm install
+
+    - name: Build ${{ matrix.name }}
+      working-directory: downstream
+      run: jlpm build
```
krassowski, line 61: This is great, I wonder if we could also add some lightweight check at the endo confirming that the built artifacts actually exist & upload them as an artifact for inspection - what do you think?

.github/workflows/downstream-tests.yml
Darshan808: CI failed [here](https://github.com/jupyterlab/jupyter-builder/actions/runs/25209849978/job/73917938350) because `hatchling` is not available in the runtime environment when `npm run build` executes. 

Currently, `hatchling` is only declared under `[build-system].requires` in both `jupyter-builder` and `jupyterlab`. This means it is installed only in isolated build environments, but it is not available during runtime and when running these Node-based build steps.

The `integrity` build script.
```ts
"integrity": "node scripts/ensure-buildutils.js && node buildutils/lib/ensure-repo.js",
```
eventually calls [getPythonVersion()](https://github.com/jupyterlab/jupyterlab/blob/ea67b8216e94d3711eb5b2556573c3fbebd47fab/buildutils/src/utils.ts#L212-L219) which executes `hatchling version`

And I think this introduces a runtime dependency on the hatchling CLI, which is not guaranteed to be installed leading to the CI failure.

From what I can tell, hatchling is only being used here to extract the JupyterLab version.

```ts
/**
 * Get the current version of JupyterLab
 */
export function getPythonVersion(): string {
  const cmd = 'hatchling version';
  const lines = run(cmd, { stdio: 'pipe' }, true).split('\n');
  return lines[lines.length - 1];
}
```

The `notebook` build passes without `hatchling`.

.github/workflows/downstream-tests.yml
Darshan808: That sounds good.
So do we build a wheel for the projects and upload it ?

krassowski: Well, I was thinking just the `jlpm build` outputs, but a full wheel could be nice too as it would make it easy to actually test the thing locally. I am not sure how much complexity and time on CI it would add though, so my answer for the wheel idea would be "maybe".

Ultimately, what is interesting (beside the ability to manually test the output), is the difference between the build artifacts for before and after the commit/PR we are looking at. I don't see a good way to compare them on CI so maybe it is ok to just upload. One easy statistic we could log in the run output is the size of the artifacts.

Darshan808: So we have 16.2 MB jupyterlab wheel and 14.7 MB notebook wheel uploaded [here](https://github.com/jupyterlab/jupyter-builder/actions/runs/25248533147/job/74036794625)

Darshan808: And the extra time for building and uploading wheels isn't significant. In `jupyterlab` it took about 30 seconds extra on CI on `notebook` it took 12 seconds extra.

krassowski (approved):
Thank you, let's try it!

PR #94, opened 2026-05-02
Fix CI: remove duplicate jobs (check release and check link), detect typescript build issues early

Darshan808:
### Description

There was an issue with typescript package build but all build tests were passing in #93 and only `check release` was failing on CI. We should build `@jupyter/builder` too in `Build` workflow.


Darshan808:
Should we rename this other job ? It is named as `build` but only deals with python package installation and pytest. Or should we also include typescript package build step here ?
https://github.com/jupyterlab/jupyter-builder/blob/a140a88fc8b2d04b648cef9c2d6712713c5a9d04/.github/workflows/python-tests.yml#L16-L37

krassowski:
Well, we definitely do not need to run "check links" twice:

https://github.com/jupyterlab/jupyter-builder/blob/a140a88fc8b2d04b648cef9c2d6712713c5a9d04/.github/workflows/python-tests.yml#L119-L126

https://github.com/jupyterlab/jupyter-builder/blob/a140a88fc8b2d04b648cef9c2d6712713c5a9d04/.github/workflows/build.yml#L69-L76

krassowski:
Maybe we could merge these two workflows and indeed rename the job you mention to just `test`?

Darshan808:
Tried running the coverage check on CI. But it seems to be too low.

> ---------------- RESULT: FAILED (minimum: 80.0%, actual: 16.4%) ----------------

Leaving it as commented for now. Will open a new issue to track it.

krassowski (approved):
Thank you!

PR #96, opened 2026-05-05
Enforce stricter linting across the entire codebase, covering both TypeScript and Python files.

### LINKED ISSUE #80: Enforce tighter linting (commented out code, top-level imports etc)
## Description

It seems like some code smells from the earlier work remain, some of which we just bumped into recently (imports not being at top-level of modules) and others including unused commented out code:

https://github.com/jupyterlab/jupyter-builder/blob/fdb7c7eb5f91e4dd46d1b0c6b29fdba7e2254ddc/jupyter_builder/federated_extensions_requirements.py#L12-L20

We should enforce better linting with `ruff` rules. Maybe we could follow whatever JupyterLab uses, or be even stricter as this is a small codebase in which it should be easy to adopt more rules.

Darshan808:
### Description

- Upgrade to ESLint v9
- Resolve all linting issues in TypeScript files
- Enabled `"ALL"` ruff rules and used an AI agent (Claude Code with Sonnet 4.6) to fix ~500 lint errors across python files.


Darshan808:
Used an AI agent to resolve the Ruff lint errors; it handled most cases well.

krassowski (commented):
I would disable `FBT001` it is just not practical here since we need to maintain compatibility, and it appears everywhere now

src/webpack-plugins.ts
```diff
@@ -14,6 +14,39 @@ import type { PluginOptions } from 'license-webpack-plugin/dist/PluginOptions';
...
+    dirTimestamps: Map<string, string | number>,
+    changedFiles: Set<string>,
+    removedFiles: Set<string>
+  ) => void;
+  callbackUndelayed: (file: string, changeTime: number) => void;
+}
+
+interface IWatchFileSystem {
+  watch(props: IWatchProps): IWatcher;
+}
```
krassowski: Is it possible to import these types from somewhere? How does it differ from `rspack.WatchFileSystem`?

jupyter_builder/jupyterlab_semver.py
```diff
@@ -1,6 +1,11 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+"""Python port of the node-semver library for semantic version parsing and comparison."""
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
```
krassowski: Got duplicated, but possibly should not have been here to begin with

```suggestion
"""Python port of the node-semver library for semantic version parsing and comparison."""
```

jupyter_builder/jupyterlab_semver.py
krassowski, line 1: By the way, can we also drop https://github.com/jupyterlab/jupyterlab/blob/main/jupyterlab/semver.py from JupyterLab now? If so, can you open a PR to remove it?

src/webpack-plugins.ts
Darshan808: Yes it is possible. Thanks!

krassowski (commented):
Looks good to my human eye, I will ask copilot for a pass

src/build-labextension.ts
```diff
@@ -75,35 +75,39 @@ commander
...
         compiler.purgeInputFileSystem();
       }
 
       if (err) {
         console.error(err.stack || err);
-        if (err.details) {
-          console.error(err.details);
+        if (err.message) {
+          console.error(err.message);
         }
```
krassowski: For this change, is `message` same as `details`? How can we test it?

jupyter_builder/federated_extensions.py
```diff
@@ -332,11 +347,11 @@ def _select_builder_marker(ext_data):
     return None, None
 
 
-def _ensure_builder(ext_path, core_package_file):
-    """Ensure that we can build the extension and return the builder script path"""
-    with open(core_package_file) as fid:
+def _ensure_builder(ext_path: str, core_package_file: str) -> str:
```
krassowski: Is this correct return type annotation? If not, why was it not caught by mypy?

src/build-labextension.ts
Darshan808: I typed `err` as the standard `Error`, which does not include a `details` property. However, it seems there is a custom `RspackError` type that extends `Error` and includes `details`. Using `details` here would probably be more appropriate.
Thanks!

jupyter_builder/federated_extensions.py
Darshan808: > Is this correct return type annotation?

No, this inconsistency was introduced while resolving the merge conflicts above.

> Why was it not caught by mypy?

It seems mypy is not currently configured to run in CI or pre-commit hooks. I'm configuring that now.

Darshan808:
Could you ask copilot for one more pass?

krassowski:
Also needs resolving conflicts due to rspack 2 merge

krassowski:
In general, changes in semver file are why I went for another round of copilot. These are hard to reason about and there are no unit tests for that functionality. I worry that we could break some version checking here due to subtle differences in logic flow. Could you revert the more involved changes (typing additions are fine) and just add comments asking linter to ignore specific lines in that file?

jupyter_builder/federated_extensions.py
```diff
@@ -500,16 +524,16 @@ def _get_labextension_metadata(module):  # noqa
     try:
         m = importlib.import_module(module)
         if hasattr(m, "_jupyter_labextension_paths"):
-            return m, m._jupyter_labextension_paths()
-    except Exception as exc:
+            return m, m._jupyter_labextension_paths()  # noqa: SLF001
+    except Exception as exc:  # noqa: BLE001
         errors.append(exc)
```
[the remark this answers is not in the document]
krassowski: There was no flow change here, since this is pre-existing behaviour resolving this as a false positive.

jupyter_builder/jupyterlab_semver.py
```diff
@@ -550,61 +578,79 @@ def inc(self, release, identifier=None):
...
 
     if anum and bnum:
-        a = int(a)
-        b = int(b)
+        ia, ib = int(a), int(b)
+        if ia < ib:
+            return -1
+        if ia > ib:
+            return 1
+        return 0
```
krassowski: Could you keep it as it was just add ignore comment?

jupyter_builder/jupyterlab_semver.py
```diff
@@ -1207,13 +1301,13 @@ def outside(version, range_, hilo, loose):
...
 
     #  If the lowest version comparator has an operator and our version
     #  is less than it then it isn't higher than the range
-    if (not low.operator or low.operator == comp) and ltefn(version, low.semver):  # noqa SIM114
-        return False
-    elif low.operator == ecomp and ltfn(version, low.semver):
+    if (not low.operator or low.operator == comp) and ltefn(version, low.semver, loose):
         return False
-    return True
+    return not (low.operator == ecomp and ltfn(version, low.semver, loose))
```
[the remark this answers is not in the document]
krassowski: - Should line 1305 raise instead of returning True?
- loose param was added here, is it correct?

Basically I am not sure of the expected behaviour here and changes are very subtle, probably hard to debug downstream

Darshan808:
The semver file’s [original source](https://github.com/podhmo/python-semver/blob/b42e9896e391e086b773fc621b23fa299d16b874/semver/__init__.py) has also remained unchanged for the past 9 years. Given that, I agree it’s probably safer to ignore the mypy errors here rather than risk introducing subtle logic regressions.

krassowski (approved):
Thank you!

PR #98, opened 2026-05-06
Patch to fully support `@jupyterlab/builder`

Darshan808:
### Description
Downstream extensions that download latest `jupyterlab` but do not yet depend upon `@jupyter/builder` and still only depend upon `@jupyterlab/builder` are having trouble building due to some design modification in `jupyter-builder`.
This PR adds a patch to support `@jupyterlab/builder` too.

### DIFF
--- a/jupyter_builder/federated_extensions.py
+++ b/jupyter_builder/federated_extensions.py
@@ -242,6 +242,7 @@ def build_labextension(  # noqa: PLR0913
         arguments.append("--development")
     if source_map:
         arguments.append("--source-map")
+
     subprocess.check_call(arguments, cwd=ext_path)  # noqa S603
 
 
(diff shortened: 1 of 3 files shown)


Darshan808:
We have a backward compatibility check, but it only tests for version mismatch error and does not test a successful build with `@jupyterlab/builder`.

https://github.com/jupyterlab/jupyter-builder/blob/376a5323e6fbbef6d643e316469f5bbc6a4c621f/tests/test_tpl.py#L189-L228

Darshan808:
I added same tests to #99 as well and we can verify that currently `@jupyterlab/builder` is not supported and CI fails with
> error: required option '--core-path <path>' not specified

But the same tests passes here.

krassowski (approved):
Thank you!

PR #102, opened 2026-05-17
Switch to stdlib HTTP + npm registry tarball fetching

### LINKED ISSUE #82: Prefer fetching `core.package.json` from npm over unpgk
> I'm a little sad there's mention of `core.package.json` from `unpkg` (often flagged by various firewalls for re-hosting malware) and `raw.github`, as this means the data won't be available directly from a semi-validated source like an npm registry. 

 _Originally posted by @bollwyvl in [#80](https://github.com/jupyterlab/jupyter-builder/issues/80#issuecomment-4229788939)_

This refers to:

https://github.com/jupyterlab/jupyter-builder/blob/fdb7c7eb5f91e4dd46d1b0c6b29fdba7e2254ddc/jupyter_builder/core_path.py#L121-L126

We already use npm API here:

https://github.com/jupyterlab/jupyter-builder/blob/fdb7c7eb5f91e4dd46d1b0c6b29fdba7e2254ddc/jupyter_builder/core_path.py#L79-L90

Maybe we could use npm tarball like:

```ts
import requests
import tarfile
import io
from pathlib import Path

def _download_npm_core_meta(version: str, destination: Path) -> None:
    destination.parent.mkdir(parents=True, exist_ok=True)

    package = "@jupyterlab/core-meta"
    meta_url = f"https://registry.npmjs.org/{package}/{version}"

    r = requests.get(meta_url, timeout=10)
    r.raise_for_status()
    metadata = r.json()

    tarball_url = metadata["dist"]["tarball"]

    r = requests.get(tarball_url, timeout=20)
    r.raise_for_status()

    # Step 4: Extract core.package.json
    with tarfile.open(fileobj=io.BytesIO(r.content), mode="r:gz") as tar:
        for member in tar.getmembers():
            if member.name.endswith("core.package.json"):
                f = tar.extractfile(member)
                if f:
                    destination.write_bytes(f.read())
                    return

    raise FileNotFoundError("core.package.json not found in package tarball")
```

I have not tested the code above.

Maybe we should also allow configuring a different NPM registry (via a traitlet) in case if folks are using custom proxies to build things locally.

Darshan808:
### Description

- Drop `requests` dependency and replace it with Python stdlib `urllib.request`
- Introduce `NPM_REGISTRY` constant for the registry base URL (we could make it configurable)
- Rewrite `_download_npm_core_meta` to fetch metadata via the official npm registry tarball flow: `registry.npmjs.org/<pkg>/<version>` → use `dist.tarball` → extract `core/package.json` instead of using unpkg

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,13 +127,14 @@ watch_labextension(
 
 ### Environment variables
 
-jupyter-builder supports the following environment variable to override the npm registry URL —
+jupyter-builder supports the following environment variables to override network URLs —
 for example, to point at an internal mirror or a local proxy. A warning is emitted at startup
-whenever the variable is set.
+whenever a variable is set.
 
-| Variable        | Default                      | Purpose                                                           |
-| --------------- | ---------------------------- | ----------------------------------------------------------------- |
-| `JPBLD_NPM_URL` | `https://registry.npmjs.org` | npm registry used to resolve and download `@jupyterlab/core-meta` |
+| Variable               | Default                             | Purpose                                                           |
+| ---------------------- | ----------------------------------- | ----------------------------------------------------------------- |
+| `JPBLD_NPM_URL`        | `https://registry.npmjs.org`        | npm registry used to resolve and download `@jupyterlab/core-meta` |
+| `JPBLD_RAW_GITHUB_URL` | `https://raw.githubusercontent.com` | Raw GitHub content URL used as a fallback when npm is unavailable |
 
 **Example — redirect to a corporate npm mirror:**
 
--- a/README.md
+++ b/README.md
@@ -127,15 +127,13 @@ watch_labextension(
 
 ### Environment variables
 
-jupyter-builder uses several URLs when fetching JupyterLab core metadata. Each can be overridden
-via an environment variable — for example, to point at an internal mirror or a local proxy. A
-warning is emitted at startup whenever a variable is set.
+jupyter-builder supports the following environment variable to override the npm registry URL —
+for example, to point at an internal mirror or a local proxy. A warning is emitted at startup
+whenever the variable is set.
 
-| Variable               | Default                             | Purpose                                                           |
-| ---------------------- | ----------------------------------- | ----------------------------------------------------------------- |
-| `JPBLD_NPM_URL`        | `https://registry.npmjs.org`        | npm registry used to resolve and download `@jupyterlab/core-meta` |
-| `JPBLD_GITHUB_URL`     | `https://github.com`                | Base GitHub URL (reserved for future use)                         |
-| `JPBLD_RAW_GITHUB_URL` | `https://raw.githubusercontent.com` | Raw GitHub content URL used as a fallback when npm is unavailable |
+| Variable        | Default                      | Purpose                                                           |
+| --------------- | ---------------------------- | ----------------------------------------------------------------- |
+| `JPBLD_NPM_URL` | `https://registry.npmjs.org` | npm registry used to resolve and download `@jupyterlab/core-meta` |
 
 **Example — redirect to a corporate npm mirror:**
 
--- a/README.md
+++ b/README.md
@@ -125,6 +125,33 @@ watch_labextension(
 )
 ```
 
+### Environment variables
+
+jupyter-builder uses several URLs when fetching JupyterLab core metadata. Each can be overridden
+via an environment variable — for example, to point at an internal mirror or a local proxy. A
+warning is emitted at startup whenever a variable is set.
+
+| Variable               | Default                             | Purpose                                                           |
+| ---------------------- | ----------------------------------- | ----------------------------------------------------------------- |
+| `JPBLD_NPM_URL`        | `https://registry.npmjs.org`        | npm registry used to resolve and download `@jupyterlab/core-meta` |
+| `JPBLD_GITHUB_URL`     | `https://github.com`                | Base GitHub URL (reserved for future use)                         |
+| `JPBLD_RAW_GITHUB_URL` | `https://raw.githubusercontent.com` | Raw GitHub content URL used as a fallback when npm is unavailable |
+
+**Example — redirect to a corporate npm mirror:**
+
+```bash
+export JPBLD_NPM_URL=https://npm.internal.example.com
+jupyter-builder build /path/to/extension
+```
+
+**Core metadata resolution order**
+
+When no explicit `--core-version` is given, jupyter-builder looks for
+`@jupyterlab/core-meta` in the extension's `node_modules` first (no network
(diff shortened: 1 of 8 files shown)


package.json
Darshan808, line 1: linter forced this!

package.json
```diff
@@ -1,100 +1,100 @@
...
+    "prettier": {
+        "singleQuote": true,
+        "trailingComma": "none",
+        "arrowParens": "avoid",
+        "endOfLine": "auto",
+        "overrides": [
+            {
+                "files": "package.json",
+                "options": {
+                    "tabWidth": 4
```
krassowski: Well, you could force it back to reduce diff using


```suggestion
                    "tabWidth": 2
```

It looks good to me either way (2 or 4).

jupyter_builder/core_path.py
```diff
@@ -3,19 +3,31 @@
...
 
 
 def _home_dir() -> Path:
     home = os.environ.get("HOME")
     return Path(home) if home else Path.home()
 
 
+def _http_get(url: str, *, headers: dict[str, str] | None = None, timeout: int = 10) -> bytes:
+    req = urllib.request.Request(url, headers=headers) if headers else url  # noqa: S310
+    with urllib.request.urlopen(req, timeout=timeout) as resp:  # noqa: S310
```
krassowski: Per https://docs.astral.sh/ruff/rules/suspicious-url-open-usage/ we could just a guard here and drop `noqa`

```suggestion
    if not url.startswith(("http:", "https:")):
        raise ValueError("URL must start with 'http:' or 'https:'")
    req = urllib.request.Request(url, headers=headers) if headers else url
    with urllib.request.urlopen(req, timeout=timeout) as resp:
```

As of today, this should never raise the `ValueError`, but it does not hurt to be extra careful

krassowski (commented):
Just some minor suggestions, will ask copilot to take a look too

jupyter_builder/core_path.py
```diff
@@ -3,19 +3,37 @@
...
 
+def _http_get(url: str, *, headers: dict[str, str] | None = None, timeout: int = 10) -> bytes:
+    if not url.startswith(("http:", "https:")):
+        msg = "URL must start with 'http:' or 'https:'"
+        raise ValueError(msg)
+    request_headers = {"User-Agent": "jupyter-builder"}
+    if headers:
+        request_headers.update(headers)
+    req = urllib.request.Request(url, headers=request_headers)  # noqa: S310
+    with urllib.request.urlopen(req, timeout=timeout) as resp:  # noqa: S310
```
krassowski, line 35: Hmm do we stil need `noqa: S310` here? In any case, looks good!

krassowski (dismissed):
Thank you!

jupyter_builder/core_path.py
Darshan808: Yes. Ruff is still giving error.

Darshan808:
Did few changes and made `NPM_REGISTRY` configurable as discussed in the issue.

README.md
```diff
@@ -125,6 +125,33 @@ watch_labextension(
...
+### Environment variables
+
+jupyter-builder uses several URLs when fetching JupyterLab core metadata. Each can be overridden
+via an environment variable — for example, to point at an internal mirror or a local proxy. A
+warning is emitted at startup whenever a variable is set.
+
+| Variable               | Default                             | Purpose                                                           |
+| ---------------------- | ----------------------------------- | ----------------------------------------------------------------- |
+| `JPBLD_NPM_URL`        | `https://registry.npmjs.org`        | npm registry used to resolve and download `@jupyterlab/core-meta` |
+| `JPBLD_GITHUB_URL`     | `https://github.com`                | Base GitHub URL (reserved for future use)                         |
```
krassowski: I would drop this since it is not needed, I think it would be more confusing than useful to include a variable that is not used.


```suggestion
```

jupyter_builder/constants.py
```diff
@@ -0,0 +1,35 @@
...
+"""URL constants for jupyter-builder network fetches, overridable via environment variables."""
+
+import os
+import warnings
+
+#: Base URL for the npm registry
+JPBLD_NPM_URL: str = os.getenv("JPBLD_NPM_URL", "https://registry.npmjs.org")
+
+#: Base GitHub URL for the jupyterlab/jupyterlab repository
+JPBLD_GITHUB_URL: str = os.getenv("JPBLD_GITHUB_URL", "https://github.com")
```
krassowski:
```suggestion
```

jupyter_builder/constants.py
```diff
@@ -0,0 +1,35 @@
...
+        f"jupyter-builder: JPBLD_NPM_URL overridden to {JPBLD_NPM_URL!r}",
+        UserWarning,
+        stacklevel=2,
+    )
+if os.getenv("JPBLD_GITHUB_URL") is not None:
+    warnings.warn(
+        f"jupyter-builder: JPBLD_GITHUB_URL overridden to {JPBLD_GITHUB_URL!r}",
+        UserWarning,
+        stacklevel=2,
+    )
```
krassowski:
```suggestion
```

jupyter_builder/constants.py
```diff
@@ -0,0 +1,23 @@
...
+"""URL constants for jupyter-builder network fetches, overridable via environment variables."""
+
+import os
+import warnings
+
+#: Base URL for the npm registry
+JPBLD_NPM_URL: str = os.getenv("JPBLD_NPM_URL", "https://registry.npmjs.org")
+
+#: Base GitHub URL for the jupyterlab/jupyterlab repository
+JPBLD_GITHUB_URL: str = "https://github.com"
```
krassowski:
```suggestion
```

README.md
```diff
@@ -125,6 +125,31 @@ watch_labextension(
...
 
+### Environment variables
+
+jupyter-builder supports the following environment variable to override the npm registry URL —
+for example, to point at an internal mirror or a local proxy. A warning is emitted at startup
+whenever the variable is set.
+
+| Variable        | Default                      | Purpose                                                           |
+| --------------- | ---------------------------- | ----------------------------------------------------------------- |
+| `JPBLD_NPM_URL` | `https://registry.npmjs.org` | npm registry used to resolve and download `@jupyterlab/core-meta` |
```
krassowski: `JPBLD_RAW_GITHUB_URL` is not here anymore?

Darshan808: I assumed that variable was also never going to be changed by any user. So removed it.
Should we include it?

Darshan808:
Tests will pass after we merge https://github.com/jupyterlab/jupyter-builder/pull/103 and rebase

krassowski (approved):
Thank you!

PR #103, opened 2026-05-20
Bump `@jupyterlab/core-meta`

Darshan808:
### Description

This was must before a stable v1 release. So that `@jupyterlab/core-meta` contains `@jupyter/builder v1`

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
       {
         "files": "package.json",
         "options": {
-          "tabWidth": 4
+          "tabWidth": 2
         }
       }
     ]
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "watch": "tsc -w --sourceMap"
   },
   "dependencies": {
-    "@jupyterlab/core-meta": "4.6.0-alpha.4",
+    "@jupyterlab/core-meta": "4.6.0-beta.0",
     "@module-federation/runtime-tools": "^2.0.0",
     "@rspack/core": "^2.0.2",
     "ajv": "^8.12.0",
--- a/tests/test_tpl.py
+++ b/tests/test_tpl.py
@@ -270,9 +270,9 @@ def test_builder_version_mismatch(tmp_path):
 
     package_json_path = extension_folder / "package.json"
 
-    # Modify the @jupyterlab/builder version to an incompatible range
+    # Modify the @jupyter/builder version to an incompatible range
     package_data = json.loads(package_json_path.read_text())
-    package_data["devDependencies"]["@jupyterlab/builder"] = "3.0.0"
+    package_data["devDependencies"]["@jupyter/builder"] = "0.0.9"
     package_json_path.write_text(json.dumps(package_data, indent=2))
 
     env = os.environ.copy()
@@ -299,7 +299,7 @@ def test_builder_version_mismatch(tmp_path):
     assert re.search(
         (
             r"ValueError: Extensions require a devDependency on @jupyterlab/builder@\^.+?, "
-            r"you have a dependency on 3\.0\.0"
+            r"you have a dependency on 0\.0\.9"
         ),
         excinfo.value.stderr,
     ), "Expected version mismatch error message not found in output!"
--- a/tests/test_tpl.py
+++ b/tests/test_tpl.py
@@ -270,9 +270,9 @@ def test_builder_version_mismatch(tmp_path):
 
     package_json_path = extension_folder / "package.json"
 
-    # Modify the @jupyter/builder version to an incompatible range
+    # Modify the @jupyterlab/builder version to an incompatible range
     package_data = json.loads(package_json_path.read_text())
-    package_data["devDependencies"]["@jupyter/builder"] = "0.0.9"
+    package_data["devDependencies"]["@jupyterlab/builder"] = "4.0.0"
     package_json_path.write_text(json.dumps(package_data, indent=2))
 
     env = os.environ.copy()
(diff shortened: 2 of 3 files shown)


Darshan808:
Hmm, so `jupyter-builder build .` is failing in extensions which depend only on `@jupyterlab/builder`.
It seems like its because extensions pull in `license-webpack-plugin@^2.3.14` (via `@jupyterlab/builder`), which has a bug in `getActualFilename`.

With webpack 5.107, some module identifiers (e.g. `ProvideSharedModule`) don’t contain `=`, but the plugin still does `filename.split('=')[1].trim()`, which crashes when the split result is `undefined`.

But this can be reproduced without `jupyter-builder`:
1. Fresh extension → `jlpm` + `jlpm build:prod` → same error  
2. Delete `yarn.lock` and `node_modules` in an existing extension → reinstall + `jlpm build:prod` → same error  

The lockfile has older webpack (5.98.0) in old extensions which doesn’t hit this code path, but freshly resolved 5.107.0 does.

So this CI failure doesn't seem to be related.

Darshan808:
@krassowski

> But this can be reproduced without jupyter-builder:

Could you help confirm this issue?

There’s no issue with `@jupyter/builder` because it pins `license-webpack-plugin@^4.0.2`, where the bug is already fixed. The problem is with `@jupyterlab/builder`, which still pins `license-webpack-plugin@^2.3.14`, where the issue exists.

I’m not sure what the best fix is here. Maybe raise an issue on https://github.com/webpack/webpack ?

Darshan808:
We can see that CI has started failing on https://github.com/jupyterlab/extension-template: https://github.com/jupyterlab/extension-template/actions/runs/26198995338/job/77084621851

Darshan808:
Seems like it's due to https://github.com/webpack/webpack/pull/20860
```ts
  // Before (≤ 5.106)
  `provide module (${shareScope}) ${name}@${version} = ${request}`
  // → "provide module (default) react@^18.0.0 = ./node_modules/react/index.js"
```
```ts
  // After (5.107)
  `provide module (${shareScope}) ${name}@${version}|${request}`
  // → "provide module (default) react@^18.0.0|./node_modules/react/index.js"
```
The separator changed from `=` to `|`.
`license-webpack-plugin 2.x` does `filename.split('=')[1].trim()`, which now gets `undefined` and crashes.

Darshan808:
The only test that's  failing on CI is
```
FAILED tests/test_tpl.py::test_files_build_jupyterlab_builder
```
where we use `jupyter-builder` with `@jupyterlab/builder` (older) to build extension.
It has nothing to do with either `jupyter-builder` or `@jupyter/builder`.

The build fails on any extension that depends upon `@jupyterlab/builder` as a I described above
> But this can be reproduced without jupyter-builder:
Fresh extension → jlpm + jlpm build:prod → same error
Delete yarn.lock and node_modules in an existing extension → reinstall + jlpm build:prod → same error

And also seen in this action: https://github.com/jupyterlab/extension-template/actions/runs/26198995338/job/77084621851

So I think this PR is good to merge as it is not related at all.

krassowski:
Can we pin `webpack` down to `5.106` on CI for the relevant tests for now just to have green CI (i.e. to make sure no regressions slip through)?

I think bumping in `license-webpack-plugin` in `@jupyterlab/builder` (`4.5.x` branch of JupyterLab) sounds good.

Separately we need to investigate if this is the right plugin to depend on in this project.

Darshan808:
There are a few things happening here that are causing the test to pass, even though the extension build should actually fail. Documenting this here for others (and for myself) for future reference.

The root cause is that the newly released `@jupyterlab/core-meta@4.6.0-beta.0` on npm no longer includes `@jupyterlab/builder`.

The current test flow is:

1. Copy the extension template.
2. Install dependencies, which installs `@jupyterlab/builder`.
3. Build the extension.
4. During the build, detect that the extension depends on `@jupyterlab/builder`, but not on `@jupyterlab/core-meta` (as this is only pulled when extension depends upon `@jupyter/builder`).
5. Fetch the latest `@jupyterlab/core-meta` from npm and use it to verify compatibility.

Previously, this worked because the latest `@jupyterlab/core-meta` included `@jupyterlab/builder`, allowing compatibility checks to run correctly. That also allowed us to simulate a failing compatibility test by modifying the `@jupyterlab/builder` version in the extension’s `package.json`.

However, now that `@jupyterlab/core-meta@4.6.0-beta.0` no longer contains `@jupyterlab/builder`, the compatibility check is effectively skipped. As a result, the test that was supposed to fail now passes, simply because there is no `@jupyterlab/builder` version available in the fetched `core-meta` to validate against.

To restore the intended test behavior, I updated the test to use `--core-version 4.5.x` when building. This causes the build process to fetch `@jupyterlab/core-meta@4.5.x` from npm instead, whose `core_package.json` still includes `@jupyterlab/builder`. That allows compatibility verification to run again and correctly reproduces the expected compatibility failure.

Once the extension template migrates to `@jupyter/builder` and drops `@jupyterlab/builder`, we will need to revisit this test (and likely a few related ones) again.

krassowski:
> To restore the intended test behavior, I updated the test to use --core-version 4.5.x when building. This causes the build process to fetch @jupyterlab/core-meta@4.5.x from npm instead, whose core_package.json still includes @jupyterlab/builder. That allows compatibility verification to run again and correctly reproduces the expected compatibility failure.
>
> Once the extension template migrates to @jupyter/builder and drops @jupyterlab/builder, we will need to revisit this test (and likely a few related ones) again.

Can you open issue(s) to track this?

krassowski (approved):
Thank you!

PR #108, opened 2026-05-23
Remove `setuptools` dependency

### LINKED ISSUE #106: Remove setuptools dependency
### Problem

`setuptools` is a bit of a moving target these days, and some packages are stuck for various versions.

### Proposed Solution

Replace use of `setuptools` in `federated_extensions`, etc. with `importlib.metadata` primitives.

### Additional context

...

Darshan808:
### Description
Drop the use of `setuptools` and create custom implementation of `find_packages` and `find_namespace_packages`.


jupyter_builder/federated_extensions.py
```diff
@@ -506,6 +506,29 @@ def _get_labextension_dir(
     return labext
 
 
+def _find_packages(path: str) -> list[str]:
+    """Find all Python packages (directories containing __init__.py) under path."""
+    path_obj = Path(path)
+    return [
```
bollwyvl: It looks like the status quo `setuptools` implementation knows how to ignore a couple things like `__pycache__`, etc. we likely need some of that logic: https://github.com/pypa/setuptools/blob/v82.0.1/setuptools/discovery.py#L137

Darshan808: Ah yes. Added logic to skip `__pycache__` and dotted dir names just like in setuptools' PackageFinder.

Darshan808:
The only place we might use setuptools is in this script, which is a fallback for getting package name
https://github.com/jupyterlab/jupyter-builder/blob/9b08563eb2a36dd373c057387026f551dfcf6a75/jupyter_builder/federated_extensions.py#L548-L564

krassowski:
> The only place we might use setuptools is in this script, which is a fallback for getting package name

I think it is ok to keep, but let's change the error message so that it does not suggest adding `setup.py`. Maybe:

```
             f"The Python package `{module}` is not a valid package, " 
             "it does not specify a `name` in `pyproject.toml` nor has a legacy `setup.py` file."

**Review by krassowski — APPROVED** [review]
Thanks!

**Comment by krassowski** [comment]
Is the CI failure related to my change?

**Comment by bollwyvl** [comment]
PyPI is having a bad day.

Also adding `--disable-pip-version-check` (or setting `PIP_DISABLE_PIP_VERSION_CHECK`) when upgrading `pip` can cut down a lot on the noise.

**Comment by krassowski** [comment]
Yes but besides it, the tests which run:

```
FAILED tests/test_tpl.py::test_builder_version_mismatch - AssertionError: Expected version mismatch error message not found in output!
assert None
 +  where None = <function search at 0x1010be980>('ValueError: Extensions require a devDependency on @jupyterlab/builder@\\^.+?, you have a dependency on 4\\.0\\.0', 'Building extension in /private/var/folders/8d/778wjbv96mq1760tv6gk374m0000gn/T/pytest-of-runner/pytest-0/test_builder...g file for details: jupyterlab-debug.log\n  warnings.warn(f"See the log file for details: {log_path}", stacklevel=1)\n')
 +    where <function search at 0x1010be980> = re.search
 +    and   'Building extension in /private/var/folders/8d/778wjbv96mq1760tv6gk374m0000gn/T/pytest-of-runner/pytest-0/test_builder...g file for details: jupyterlab-debug.log\n  warnings.warn(f"See the log file for details: {log_path}", stacklevel=1)\n' = CalledProcessError(1, ['jupyter-builder', 'build', '/private/var/folders/8d/778wjbv96mq1760tv6gk374m0000gn/T/pytest-of-runner/pytest-0/test_builder_version_mismatch0/ext', '--core-version', '4.5.x']).stderr
 +      where CalledProcessError(1, ['jupyter-builder', 'build', '/private/var/folders/8d/778wjbv96mq1760tv6gk374m0000gn/T/pytest-of-runner/pytest-0/test_builder_version_mismatch0/ext', '--core-version', '4.5.x']) = <ExceptionInfo CalledProcessError(1, ['jupyter-builder', 'build', '/private/var/folders/8d/778wjbv96mq1760tv6gk374m0000gn/T/pytest-of-runner/pytest-0/test_builder_version_mismatch0/ext', '--core-version', '4.5.x']) tblen=2>.value
```

**Comment by bollwyvl** [comment]
Ah, didn't get that far. Perhaps try:

```
pytest -vv --tb=long --color=yes
```

... for more glorious technicolor :heavy_multiplication_x: 

Also, for debugging CI stuff, `pytest-html` (and archiving it `always()`) helps a lot.

Darshan808:
#115 fixes the CI

PR #109, opened 2026-05-24
Bundle `yarn.js` with `workspaces foreach` support

### LINKED ISSUE #104: Enable yarn workspaces foreach
### Problem

For downstream monorepos (such as jupyterlab, notebook, or jupyterlite) `yarn workspace foreach` is a useful  tool for avoiding heavy dependencies like `lerna`.

### Proposed Solution

- update to a version of `yarn` that includes `yarn workspaces foreach`
- rebuild `yarn.js` and bundle the appropriate plugin
  - may have additional benefits, such as more accurate license information

### Additional context

- https://github.com/jupyterlite/jupyterlite/pull/1843

Darshan808:
### Description
Builds a custom `yarn.js` (v3.5.0) with `@yarnpkg/plugin-workspace-tools` baked in, so `yarn workspaces foreach` works out of the box in downstream projects without any extra plugin setup.

### How `yarn.js` was generated

- Checked out the `@yarnpkg/cli/3.5.0` branch from the Yarn Berry repo
- Added `@yarnpkg/plugin-workspace-tools` to `packages/yarnpkg-cli/package.json` (dependency + standard bundle profile)
- Built the CLI with `yarn workspace @yarnpkg/cli build:cli --no-git-hash`
- Replaced `jupyter_builder/yarn.js` with the generated artifact

### Verification
Added a small step in CI in Downstream Tests to check whether `yarn workspace foreach <some_command>` works or not.


Darshan808:
[Before CI](https://github.com/jupyterlab/jupyter-builder/actions/runs/26352347509/job/77572717198?pr=109) confirms that the `yarn workspace foreach` command isn't present currently.

Darshan808:
And [After CI](https://github.com/jupyterlab/jupyter-builder/actions/runs/26354681569/job/77579127659?pr=109) confirms that the `yarn workspace foreach` command works on downstream projects now.

jupyter_builder/yarn.js
krassowski, line 1: Should we have a script on CI to recreate this blob and to validate that it is as expected.

bollwyvl: Further, this is basically a new package at this point, e.g `@jupyter/jlpm` and we should probably start publishing it, along with the extracted license information for vendored packages. Then all the sha stuff on #110 would be tracked canonically in `package.json` and `yarn.lock`.

Darshan808:
@bollwyvl  On the `@jupyter/jlpm` extraction discussion, should that be an npm package or a Python package?

My understanding is that it should be an npm package, since `@jupyter/` is already an npm scope and the actual payload (`yarn.js`) is JavaScript. In that model, a user could simply run:

```bash
npm install -g @jupyter/jlpm
```

and get the `jlpm` binary without needing Python at all.

What I'm less clear about is the integration with `jupyter-builder`. We want `jlpm` to be available automatically when someone installs `jupyter-builder` via:

```bash
pip install jupyter-builder
```

How would that work if `jlpm` exists purely as an npm package?

Would we still need a small Python wrapper/package that vendors or downloads `yarn.js` so that the `jlpm` CLI becomes available as part of the Python installation? In other words, would `pip install jupyter-builder` somehow need to bring in the npm package under the hood?

I'm a bit unsure about the packaging story here and how the pieces are expected to fit together if we publish `yarn.js` as a separate `@jupyter/jlpm` npm package.

Could you please provide a bit more context on the intended approach so I can look into it further ?

bollwyvl:
Right in one approach; `npm install @jupyter/jlpm` would indeed get a well-structured `node_modules/@jupyter/jlpm/dist/yarn.js`. It _could_ ship a `node_modules/.bin/jlpm`, and then _could_ then be used for e.g. `packageManager` in `package.json` (unless that's a "blessed" list we don't want to mess with, because `corepack` is crazy). Such a `jlpm` would be usable from `package.json#/scripts/`, but would not necessarily be on `$PATH` itself, depending on _which_ `node` is presently on `$PATH`, so might add to confusion.

`jupyter-builder` (the python package) would vendor this (and the `package.json`, license files, etc.), and continue to provide the equivalent, and more reliably findable-in-a-Jupyter environment, `{sys.prefix}/bin/jlpm`.

There _might_ be a compelling reason for `jlpm` itself to be a wholly standalone PyPI package, such that `jupyter-builder` contained no `.js`, but having an even partially relaxed  `jupyter-builder -> jlpm` dependency would introduce a possible "bad dimension of the matrix" situation which `jupyter-builder` is trying to avoid.

Darshan808:
Should we track the "seperation of `jlpm` into a seperate package" in a new issue and merge this one?

.github/workflows/downstream-tests.yml
```diff
@@ -56,6 +56,12 @@ jobs:
       working-directory: downstream
       run: jlpm install
 
+    - name: Test `jlpm workspaces foreach` command
+      working-directory: downstream
+      continue-on-error: true
+      run: |
+        jlpm workspaces foreach --parallel exec sh -c 'echo Hello from $PWD'
+
```
Darshan808, line 64: We should remove this check before merging!

krassowski: Well maybe it is fine to keep?

.github/workflows/downstream-tests.yml
```diff
@@ -56,6 +56,12 @@ jobs:
       working-directory: downstream
       run: jlpm install
 
+    - name: Test `jlpm workspaces foreach` command
+      working-directory: downstream
+      continue-on-error: true
+      run: |
```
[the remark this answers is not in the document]
Darshan808: Actually it even fails now in `jupyterlab` (See [CI Logs](https://github.com/jupyterlab/jupyter-builder/actions/runs/26690077088/job/78664821656?pr=109)). We'll change this once we fix it upstream.

.github/workflows/downstream-tests.yml
Darshan808: Okay, maybe it's fine to keep.

krassowski (approved):

PR #110, opened 2026-05-24
Expose `YARN_VERSION` and `YARN_SHA256`

### LINKED ISSUE #105: Expose version of vendored yarn
### Problem

Downstreams like `jupyterlab` and `notebook` need to declare their effective version of `"packageManager'; yarn@XXX` to integrate properly with other tools like `dependabot`. 

### Proposed Solution

Expose a top-level `YARN_VERSION` in the python and js side of the house.

Bonus: `YARN_SHA256`, with a test of the built wheel to validate this. Runtime validation is probably a bridge too far.

### Additional context

If exposed, this could be used by downstream linters and tests.

See also #104

Darshan808:
### Description

Expose Yarn version from the vendored `yarn.js` at import time and also compute and expose its SHA256 hash.

Currently, the generated hash matches the upstream Yarn 3.5.0 release. Once PR #109 is merged, the bundled `yarn.js` will diverge, and the hash will become specific to Jupyter Builder rather than any standard Yarn release.


jupyter_builder/_yarn_info.py
```diff
@@ -0,0 +1,25 @@
...
+
+# Minified yarn berry embeds the version as: YarnVersion:()=>VarName ... var VarName="x.y.z"
+_match = re.search(rb"YarnVersion:\(\)=>(\w+)", _yarn_bytes)
+if _match:
+    _var = re.escape(_match.group(1))
+    _match = re.search(rb"var " + _var + rb'="([^"]+)"', _yarn_bytes)
+
+if not _match:
+    msg = "Could not extract YARN_VERSION from vendored yarn.js"
+    raise RuntimeError(msg)
```
krassowski, line 22: This looks a bit fragile, I think a minimum countermeasure would be adding a test case for this.

jupyter_builder/_yarn_info.py
```diff
@@ -0,0 +1,25 @@
...
+"""Yarn version and integrity information for the vendored yarn.js."""
+
+from __future__ import annotations
+
+import hashlib
+import re
+from pathlib import Path
+
+_yarn_js = Path(__file__).resolve().parent / "yarn.js"
+_yarn_bytes = _yarn_js.read_bytes()
```
fleming79, line 12: Consider deleting this when it is no longer required as it takes significant memory to store.

jupyter_builder/_yarn_info.py
```diff
@@ -0,0 +1,25 @@
...
+_match = re.search(rb"YarnVersion:\(\)=>(\w+)", _yarn_bytes)
+if _match:
+    _var = re.escape(_match.group(1))
+    _match = re.search(rb"var " + _var + rb'="([^"]+)"', _yarn_bytes)
+
+if not _match:
+    msg = "Could not extract YARN_VERSION from vendored yarn.js"
+    raise RuntimeError(msg)
+
+YARN_VERSION: str = _match.group(1).decode()
```
fleming79, line 24: I recommend storing the version and hash as constants and comparing the hash instead with an assertion. If the hash hasn't changed, the version will be correct.

fleming79 (commented):
Some comments for consideration.

Darshan808:
I see that currently jupyterlab, notebook and other downstream extensions must use:
```json
"packageManager": "yarn@3.5.0"
```
in their `package.json` for tools like dependabot and others to function properly.

So these exported constant enables open few possibilities:

1. **jupyterlab** can extend `jlpm integrity` to treat `packageManager` as a managed field, checked against the value exported from `jupyter-builder`.
2. For extensions, I'm not entirely sure what the best approach would be. We could leave it up to extensions to add a CI test that checks they are using the expected packageManager value. Alternatively, jupyter-builder itself could validate this during the build and emit a warning such as:
```
   package.json declares "packageManager": "yarn@3.5.0", but this build uses yarn 4.9.1.
```

PR #115, opened 2026-06-11
Update template tests for new `@jupyter/builder` default

### LINKED ISSUE #107: Change tests after extension-template starts depending on v1
To restore the intended test behavior, I updated the test to use `--core-version 4.5.x` when building. This causes the build process to fetch `@jupyterlab/core-meta@4.5.x` from npm instead, whose `core_package.json` still includes `@jupyterlab/builder`. That allows compatibility verification to run again and correctly reproduces the expected compatibility failure.

Once the extension template migrates to `@jupyter/builder` and drops `@jupyterlab/builder`, we will need to revisit this test (and likely a few related ones) again.

_Originally posted by @Darshan808 in https://github.com/jupyterlab/jupyter-builder/issues/103#issuecomment-4510739445_

Darshan808:
### Description

The extension template now ships `@jupyter/builder` by default and no longer includes `@jupyterlab/builder`. Since `@jupyter/builder` is the preferred builder marker, our tests needed updating:

- Normal tests no longer adds `@jupyter/builder` explicitly — the template provides it now.
- The `*_jupyterlab_builder` tests now swap in `@jupyterlab/builder` and remove `@jupyter/builder` so they actually exercise the `jupyterlab/builder` path.

### DIFF
--- a/tests/test_tpl.py
+++ b/tests/test_tpl.py
@@ -107,9 +107,8 @@ def test_files_build_jupyterlab_builder(tmp_path):
         " p.resolutions = p.resolutions || {};"
         " p.resolutions.webpack='5.106.0';"
         " p.devDependencies = p.devDependencies || {};"
-        " const v = p.devDependencies['@jupyter/builder'] || '^4.0.0';"
         " delete p.devDependencies['@jupyter/builder'];"
-        " p.devDependencies['@jupyterlab/builder'] = v;"
+        " p.devDependencies['@jupyterlab/builder'] = '^4.0.0';"
         " fs.writeFileSync('package.json', JSON.stringify(p,null,2));"
     )
     run(["node", "-e", prepare], cwd=extension_folder, check=True)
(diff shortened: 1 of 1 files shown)


tests/test_tpl.py
```diff
@@ -270,8 +264,13 @@ def test_builder_version_mismatch(tmp_path):
 
     package_json_path = extension_folder / "package.json"
 
-    # Modify the @jupyterlab/builder version to an incompatible range
+    # The template ships `@jupyter/builder` by default, but the version
+    # incompatibility check can only be verified on `@jupyterlab/builder` for
+    # now. So we remove `@jupyter/builder` and pin `@jupyterlab/builder` to an
```
Darshan808, line 269: This is because, we don't have enough versions published yet to test for incompatibility using `@jupyter/builder`

krassowski (approved):
Thanks!

PR #122, opened 2026-06-21
fix: ignore module federation source maps to prevent build warnings

intagliated:
### Description

This PR resolves an issue where building extensions produces verbose source map parsing warnings from upstream packages (specifically `@module-federation/error-codes`).

While these builds complete successfully, missing TypeScript source files in third-party modules trigger `ENOENT: no such file or directory` errors from `source-map-loader`, which clutters the build logs.

By adding an `exclude` rule targeting `/node_modules\/@module-federation/` inside `src/extensionConfig.ts`, the builder now defensively skips parsing these known broken/missing upstream source maps.

### Related Issues

Fixes #111

### Type of Change

* [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/extensionConfig.ts
+++ b/src/extensionConfig.ts
@@ -273,7 +273,8 @@ function generateConfig({
     rules.push({
       test: /\.js$/,
       enforce: 'pre',
-      use: [require.resolve('source-map-loader')]
+      use: [require.resolve('source-map-loader')], 
+      exclude: [/node_modules\/@module-federation/]
     });
   }
 
--- a/src/extensionConfig.ts
+++ b/src/extensionConfig.ts
@@ -273,7 +273,7 @@ function generateConfig({
     rules.push({
       test: /\.js$/,
       enforce: 'pre',
-      use: [require.resolve('source-map-loader')], 
+      use: [require.resolve('source-map-loader')],
       exclude: [/node_modules\/@module-federation/]
     });
   }
--- a/tests/test_extension_config.py
+++ b/tests/test_extension_config.py
@@ -1,0 +1,69 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import re
+from pathlib import Path
+
+
+def test_source_map_loader_excludes_module_federation_in_source():
+    """Regression test for https://github.com/jupyterlab/jupyter-builder/issues/111.
+
+    @module-federation/error-codes@0.21.4 shipped source map files that
+    referenced TypeScript source files (e.g. src/error-codes.ts) which were
+    never included in the published package. This caused source-map-loader to
+    throw ENOENT warnings during extension builds.
+
+    The fix adds an exclude rule in extensionConfig.ts so that source-map-loader
+    skips the @module-federation directory entirely.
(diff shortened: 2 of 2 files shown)


Darshan808:
Thanks for working on this @intagliated 
Do you really see those logs?
Can you share your output, or even better write a test proving they exist currently.

What I see: https://github.com/jupyterlab/jupyter-builder/issues/111#issuecomment-4756311271

Darshan808:
Thanks for digging. Yes if the fix is already in the upstream and warning logs are no longer seen here, than we can close this PR.

intagliated:
Hi @Darshan808.
 **The original warning from the issue:**
 ```
 WARNING in ./node_modules/@module-federation/error-codes/dist/index.cjs.js
     ⚠ Module Warning (from source-map-loader):
     │ Failed to parse source map from '.../node_modules/@module-federation/error-codes/src/error-codes.ts'
     │ Error: ENOENT: no such file or directory
 ```

 **The dependency chain that caused it:**
 ```
 @jupyterlab/builder@4.6.0-alpha.0   ← the early test release of the builder (Dec 2025)
   → @rspack/core@1.6.4              ← the bundler it depended on at that time (version 1.6.4)
     → @module-federation/runtime-tools@0.21.4
       → @module-federation/runtime@0.21.4
         → @module-federation/error-codes@0.21.4  ← the package that caused the warning
 ```

 These are all version numbers — the number after `@` identifies the exact release of a package. `0.21.4` is an older release of `@module-federation/error-codes`, while `2.5.1` is the current stable release pulled in today. The jump from `0.x` to `2.x` indicates a major rewrite happened between these releases.

 **Why `@module-federation/error-codes@0.21.4` caused the warning:**
 Its source map file (`dist/index.cjs.js.map`) pointed to source files that were never included in the published package:
 ```json
 "sources": [
     "../src/error-codes.ts",
     "../src/getShortErrorMsg.ts",
     "../src/desc.ts"
 ]
 ```
 So `source-map-loader` tried to find these `.ts` files, couldn't, and printed the `ENOENT` warning.

 **Why `@module-federation/error-codes@2.5.1` no longer causes the warning:**
 The newer release now embeds the source content directly inside the map file via `sourcesContent` — so the build tool no longer needs to look up any external files on disk. Nothing goes missing, no warning.

 **Tests added:**
 I've also added regression tests in `tests/test_extension_config.py` that verify:
 1. The `@module-federation` exclude rule exists in `src/extensionConfig.ts`
 2. The rule survived TypeScript compilation in `lib/extensionConfig.js`
 3. The exclude rule is placed directly alongside the `source-map-loader` rule, not just mentioned elsewhere in the file

intagliated:
Sure.

intagliated:
Hi, I  want to reopen the pull request again  to understand why the CI failed -16/21

Darshan808:
I added a extension build job in a PR and here' the CI https://github.com/jupyterlab/jupyter-builder/actions/runs/28006945873/job/82890989868?pr=126
There's no any warnings as mentioned in #111. I believe we should close this PR as well as the issue.

intagliated:
ok

PR #131, opened 2026-06-27
Bump rspack to 2.1

### LINKED ISSUE #129: Update to rspack 2.1.0 and set up dependabot for rspack?
It looks like it gets circa 10% performance boost and enables TypeScript 7 type checks. Also, it looks like rslint now supports ESLint plugins.

- https://rspack.rs/blog/announcing-2-1
- https://github.com/web-infra-dev/rspack/relases/tag/v2.1.0

Darshan808:
### Description
Rspack 2.1 is out, so bumping `@rspack/core` from 2.0.2 → 2.1.1.

Also added `@rspack/core` to the dependabot allow-list so we get future rspack bumps automatically.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,3 +26,4 @@ updates:
     versioning-strategy: increase
     allow:
       - dependency-name: "@jupyterlab/core-meta"
+      - dependency-name: "@rspack/core"
--- a/.github/workflows/downstream-tests.yml
+++ b/.github/workflows/downstream-tests.yml
@@ -54,7 +54,7 @@ jobs:
 
     - name: Install deps
       working-directory: downstream
-      run: jlpm install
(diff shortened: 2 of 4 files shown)


.github/dependabot.yml
```diff
@@ -26,3 +26,4 @@ updates:
     versioning-strategy: increase
     allow:
       - dependency-name: "@jupyterlab/core-meta"
+      - dependency-name: "@rspack/core"
```
krassowski, line 29: Maybe also `@module-federation/runtime-tools`? We can defer decision to another PR.

krassowski (approved):
LGTM, thanks!

.github/dependabot.yml
fleming79: Why not all dependencies as well?

PR #136, opened 2026-06-29
Upgrade glob dependency

fleming79:
Version upgrade from 7 -> 13 and removed tilde (~) versioning.
Requires node >= v20
See [glob changelog](https://classic.yarnpkg.com/en/package/glob#changelog)


Darshan808 (approved):
Thanks @fleming79

PR #147, opened 2026-07-13
Add `py.typed`

bollwyvl:
## references
- for https://github.com/jupyterlab/jupyterlab/issues/19057

## changes
- [x] add `py.typed` marker file for [PEP-561](https://peps.python.org/pep-0561/)
  - > by inspection, this is ending up in the `.tar.gz` and `.whl`; it seems like modern `hatchling` doesn't need any extra metadata :tada: 
- [x] ignore `py.typed` in `.licenserc.yaml`


krassowski (approved):
Thank you!

PR #150, opened 2026-07-17
Add deptry to CI to catch missing/transitive dependencies

### LINKED ISSUE #84: Ensure that all imports are listed in dependencies
It would have caught #83

Darshan808:
### Description

Wired `deptry` into the existing `test_lint` job

`scripts/` is excluded from scanning, so the `click`/`packaging`/`jupyter_releaser` imports in `bump_version.py` aren't flagged.

suppressed the remaining CLI-only/plugin tools (`ruff`, `mypy`, etc) via `per_rule_ignores` for DEP002 only, since they're invoked via subprocess/entry points and never imported.

No first-party import misfired.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,10 +47,11 @@ test = [
     "pytest-check-links>=0.7",
     "pytest-cov",
     "copier>=9.3,<10",
-    "jinja2-time"
+    "jinja2-time",
+    "deptry"
 ]
 # Check ruff version is aligned with the one in .pre-commit-config.yaml
-dev = ["build", "mypy", "pre-commit", "hatch", "ruff==0.15.20", "deptry"]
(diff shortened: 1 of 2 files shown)


Darshan808:
CI shows:
> Run deptry .
> + deptry .
> Scanning 21 files...
>
> Success! No dependency issues found.

pyproject.toml
```diff
@@ -50,7 +50,7 @@ test = [
     "jinja2-time"
 ]
 # Check ruff version is aligned with the one in .pre-commit-config.yaml
-dev = ["build", "mypy", "pre-commit", "hatch", "ruff==0.15.20"]
+dev = ["build", "mypy", "pre-commit", "hatch", "ruff==0.15.20", "deptry"]
```
krassowski: Can this go to test dependencies instead?

krassowski (approved):
Thank you!

PR #151, opened 2026-07-17
Fix flaky `test_builder_version_mismatch`

### LINKED ISSUE #148: Tests failing on CI?
```
FAILED tests/test_tpl.py::test_builder_version_mismatch - AssertionError: Expected version mismatch error message not found in output!
assert None
 +  where None = <function search at 0x10397e980>('ValueError: Extensions require a devDependency on @jupyterlab/builder@\\^[^,]+, you have a dependency on 4\\.0\\.0', '/Users/runner/work/jupyter-builder/jupyter-builder/jupyter_builder/debug_log_file_mixin.py:62: UserWarning: An error ...g file for details: jupyterlab-debug.log\n  warnings.warn(f"See the log file for details: {log_path}", stacklevel=1)\n')
```

Darshan808:
### Description

This test kept failing with HTTP 403: rate limit exceeded because `--core-version 4.5.x` isn't on npm, so it fell back to downloading `core-meta` from GitHub, which is rate-limited in CI.

`get_core_meta` checks the local cache before hitting the network, so the test now seeds a minimal dummy `core.package.json` for `4.5.x` before the build. It declares an incompatible `@jupyterlab/builder` version range (same as in real core meta file for `4.5.x`), which is all the test needs to trigger the version-mismatch error. This avoids all network access and eliminates the flake.

### DIFF
--- a/tests/test_tpl.py
+++ b/tests/test_tpl.py
@@ -257,10 +257,37 @@ def test_watch_functionality_jupyterlab_builder(tmp_path):
             watch_process.kill()
 
 
+def _seed_core_meta_cache(version, builder_range):
+    """Seed the core-meta cache so ``get_core_meta`` resolves offline."""
+    home = os.environ.get("HOME") or str(Path.home())
+    cache_file = (
+        Path(home) / ".cache" / "jupyterlab_builder" / "core" / version / "core.package.json"
+    )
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thank you!

PR #153, opened 2026-07-28
Improve fallback resolution

### LINKED ISSUE #152: `core-metadata` fallback needs improvements (should point to `stable` not `main`, should handle `@jupyterlab/builder` specially)
The upgrade path can lead to quite a bad outcome with `getter is not a function` or `Type 'X' does not satisfy the constraint 'Y'` errors:
- https://github.com/jupyterlab/jupyter-collaboration/pull/606
- https://github.com/jupyterlab/extension-template/issues/151

Somewhat related to https://github.com/jupyterlab/jupyter-builder/issues/145.

LLM summary of jupyter-collaboration issue:

> The branch changed the pip pin from jupyterlab>=4.6.0b1 to jupyterlab>=4.6.0, which makes CI install stable JupyterLab 4.6.2 for the first time — and the labextension build toolchain this repo uses silently builds bundles that are only compatible with JupyterLab 4.7.0-alpha. Main "passes" only because its prerelease pin makes pip install 4.7.0a1.
> 
> #### The causal chain
> The stale builder. `docprovider-extension/package.json` devDepend on @jupyterlab/builder ^4.6.0-alpha.5. No stable 4.6 of that package exists — JupyterLab replaced it with @jupyter/builder mid-4.6-cycle (jupyterlab#18723), so the lock is stuck on the alpha forever.
> 
> 1. The silent network fallback. Since JupyterLab 4.6, jupyter labextension build delegates to the jupyter-builder Python package. To compute module-federation metadata it looks for @jupyterlab/core-meta in node_modules — present only with the new @jupyter/builder. Not finding it, it warns (swallowed by pip's quiet build output) and downloads core metadata from jupyterlab's main branch — currently the 4.7.0-alpha line.
> 
> 2. The proof. I extracted the federation consume table from the CI-built bundles: they require @jupyterlab/notebook ^4.7.0-alpha.1, @jupyterlab/apputils ^4.8.0-alpha.1, @jupyterlab/services ^7.7.0-alpha.1 — exactly jupyterlab main's dev_mode resolutions, not your ^4.6.0 ranges from package.json/yarn.lock.
> 
> 3. The crash. The 4.6.2 host registers notebook 4.6.2 / apputils 4.7.2 / services 7.6.2, which don't satisfy those alpha ranges. The Module-Federation runtime then loads the extension's own bundled copies; where no bundled fallback exists it calls the undefined fallback getter — that's literally the TypeError: getter is not a function. The duplicated @jupyterlab/notebook means jupyter-docprovider's executor plugin (executor.ts) provides a different INotebookCellExecutor token object than core expects — and since docprovider disables core's cell-executor plugin, core's tracker gets "No provider for: INotebookCellExecutor" and takes ~25 dependent plugins down with it. That's your whole log.
> 
> 4. Why other jobs mislead you. ui_tests passes because the workflow hardcodes pip install "jupyterlab>=4.6.0b1" → it still tests against 4.7.0a1. Windows "passes" because browser_check there never actually launches a browser (finishes in 3 s with "Test Complete"). The windows-3.12 and main's ubuntu-3.11 reds are unrelated pytest failures.

What we could do:
- a) Stop defaulting to `main` in the fallback - use `latest` release instead
- b) After resolving core metadata, if jupyterlab is importable and its `major.minor` disagrees with the core-meta version, fail with a one-line message ("building against 4.7.0-alpha metadata but jupyterlab 4.6.2 is installed") - but only if the core-path or version was not given explicitly; this is we could still support cross-compilation, but IF we are using a fallback we could detect bad fallback resolution
- c) Treat the legacy `@jupyterlab/builder` marker as a known situation, not a coincidence. jupyter-builder already detects the old marker, so it knows it's building a pre-migration extension - and currently proceeds to combine an alpha-era rspack-1 builder with dev-branch metadata. It could instead resolve core-meta to the old builder's own version (faithful to old behavior), or refuse with migration instructions. This is basically follow-up on #81.

Darshan808:
### Summary

This PR makes three small improvements to how `get_core_meta` selects a JupyterLab version when the caller does not explicitly provide one.

1. **Use the latest stable release instead of `main` by default**
   - It now resolves `"latest"` using npm's `latest` dist-tag, with GitHub's `releases/latest` API as a fallback.

2. **Fail fast when fallback version resolution is incompatible**
   - When no version, `ext_path`, or `core-meta` is provided and `get_core_meta` has to infer a version, it now validates the resolved version against the installed JupyterLab (if present).
   - If their `major.minor` versions do not match, the build now fails immediately instead of proceeding with incompatible metadata. For example, I got this when I removed `jupyterlab/core-meta` from node modules and initiated the build in the environment where I have older version of juptyerlab:
   

    Screenshot 2026-07-28 at 13 31 43
    
   - This check only applies to inferred versions. Explicitly requested versions remain unchanged, so cross-compiling against a different JupyterLab version is still supported.

3. **Resolve legacy `@jupyterlab/builder` extensions correctly**
   - Now `get_core_meta` resolves the core metadata version from the extension's pinned `@jupyterlab/builder` dependency instead of using "latest" core metadata.
   - A warning is emitted explaining the behaviour and encouraging migration to `@jupyter/builder`.

### DIFF
--- a/tests/test_core_path.py
+++ b/tests/test_core_path.py
@@ -471,6 +471,89 @@ def test_get_core_meta_latest_uses_npm_latest_and_caches_by_resolved_version(tmp
     )
 
 
+def test_get_core_meta_defaults_to_latest_when_not_installed(tmp_path, monkeypatch):
+    """With no explicit version and no installed core-meta, the npm "latest" dist-tag is used."""
+    ext_path = tmp_path / "ext"
+    ext_path.mkdir()
+    (ext_path / "node_modules").mkdir()
+    monkeypatch.setenv("HOME", str(tmp_path))
+
+    tarball_content = b'{"devDependencies": {}}'
+    tarball_bytes = _make_core_package_tarball(tarball_content)
+    tarball_url = f"{core_path.JPBLD_NPM_URL}/@jupyterlab/core-meta/-/core-meta-4.5.7.tgz"
+    registry_meta = json.dumps({"dist": {"tarball": tarball_url}}).encode()
+
+    calls = []
+
+    def fake_urlopen(req_or_url, **_kwargs):
+        url = getattr(req_or_url, "full_url", req_or_url)
+        calls.append(url)
+        if url == f"{core_path.JPBLD_NPM_URL}/@jupyterlab/core-meta/latest":
+            return io.BytesIO(json.dumps({"version": "4.5.7"}).encode())
+        if url == f"{core_path.JPBLD_NPM_URL}/@jupyterlab/core-meta/4.5.7":
+            return io.BytesIO(registry_meta)
+        if url == tarball_url:
+            return io.BytesIO(tarball_bytes)
+        msg = f"Unexpected URL {url}"
+        raise AssertionError(msg)
+
+    monkeypatch.setattr(core_path.urllib.request, "urlopen", fake_urlopen)
+
+    location = core_path.get_core_meta(ext_path=ext_path)
+
+    assert calls == [
+        f"{core_path.JPBLD_NPM_URL}/@jupyterlab/core-meta/latest",
+        f"{core_path.JPBLD_NPM_URL}/@jupyterlab/core-meta/4.5.7",
+        tarball_url,
+    ]
+    assert location == str(
+        tmp_path / ".cache" / "jupyterlab_builder" / "core" / "4.5.7" / "core.package.json",
+    )
+
+
+def test_get_core_meta_latest_falls_back_to_github_when_npm_unreachable(tmp_path, monkeypatch):
+    """If npm is unreachable, "latest" is resolved to the newest GitHub release tag."""
+    ext_path = tmp_path / "ext"
+    ext_path.mkdir()
+    (ext_path / "node_modules").mkdir()
+    monkeypatch.setenv("HOME", str(tmp_path))
+
+    github_url = (
+        "https://raw.githubusercontent.com/"
+        "jupyterlab/jupyterlab/v4.5.7/"
+        "jupyterlab/staging/package.json"
+    )
+    latest_release = json.dumps({"tag_name": "v4.5.7"}).encode()
+
+    calls = []
+
+    def fake_urlopen(req_or_url, **_kwargs):
+        url = getattr(req_or_url, "full_url", req_or_url)
+        calls.append(url)
+        if url == f"{core_path.JPBLD_NPM_URL}/@jupyterlab/core-meta/latest":
+            msg = "Not Found"
+            raise urllib.error.URLError(msg)
+        if url == core_path._GITHUB_LATEST_RELEASE_API_URL:
+            return io.BytesIO(latest_release)
+        if url == github_url:
+            return io.BytesIO(b'{"dependencies": {}}')
+        msg = f"Unexpected URL {url}"
+        raise AssertionError(msg)
+
+    monkeypatch.setattr(core_path.urllib.request, "urlopen", fake_urlopen)
+
+    location = core_path.get_core_meta(ext_path=ext_path)
+
+    assert calls == [
+        f"{core_path.JPBLD_NPM_URL}/@jupyterlab/core-meta/latest",
+        core_path._GITHUB_LATEST_RELEASE_API_URL,
+        github_url,
+    ]
+    assert location == str(
+        tmp_path / ".cache" / "jupyterlab_builder" / "core" / "4.5.7" / "core.package.json",
+    )
+
+
 def test_ensure_builder_with_jupyter_builder(tmp_path):
     ext_path = tmp_path / "ext"
     core_path_dir = tmp_path / "core-meta"
--- a/tests/test_core_path.py
+++ b/tests/test_core_path.py
@@ -3,9 +3,11 @@
 
 import io
 import json
+import re
 import tarfile
 import urllib.error
 from pathlib import Path
+from unittest import mock
 
 import pytest
 
@@ -24,6 +26,19 @@ def _make_core_package_tarball(content: bytes) -> bytes:
         info.size = len(content)
         tar.addfile(info, io.BytesIO(content))
     return buf.getvalue()
+
+
+@pytest.fixture(autouse=True)
+def _no_jupyterlab_installed(monkeypatch):
+    """Default to "jupyterlab is not installed" so tests are isolated from the dev environment.
+
+    Tests that exercise the fallback-version compatibility check override this.
+    """
+
+    def _raise_not_found(_package):
+        raise core_path.PackageNotFoundError(_package)
+
+    monkeypatch.setattr(core_path, "installed_version", _raise_not_found)
 
 
 def test_get_core_meta_prefers_installed_core_meta(tmp_path):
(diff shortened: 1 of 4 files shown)


tests/test_core_path.py
```diff
@@ -471,6 +486,306 @@ def fake_urlopen(req_or_url, **_kwargs):
...
+        f"{core_path.JPBLD_NPM_URL}/@jupyterlab/core-meta/4.5.7",
+        tarball_url,
+    ]
+    assert location == str(
+        tmp_path / ".cache" / "jupyterlab_builder" / "core" / "4.5.7" / "core.package.json",
+    )
+
+
+def test_get_core_meta_latest_falls_back_to_github_when_npm_unreachable(tmp_path, monkeypatch):
+    """If npm is unreachable, "latest" is resolved to the newest GitHub release tag."""
```
krassowski, line 530: I think we should only use one or another, not both, to make it reproducible.

tests/test_core_path.py
```diff
@@ -471,6 +486,306 @@ def fake_urlopen(req_or_url, **_kwargs):
...
+    location = core_path.get_core_meta(ext_path=ext_path, logger=logger)
+
+    # The "latest" dist-tag is never hit — the marker's own version is requested directly.
+    assert calls == [
+        f"{core_path.JPBLD_NPM_URL}/@jupyterlab/core-meta/4.5.7",
+        tarball_url,
+    ]
+    assert location == str(
+        tmp_path / ".cache" / "jupyterlab_builder" / "core" / "4.5.7" / "core.package.json",
+    )
```
krassowski, line 683: I do not follow this test case. There is no `core-meta` 4.5.7 on  https://www.npmjs.com/package/@jupyterlab/core-meta?activeTab=versions (though we could back-populate it). Is this test saying that it will try to fetch it?

tests/test_core_path.py
Darshan808: Got it. Just to clarify, are you suggesting we rely solely on npm for resolving `latest` and remove the GitHub fallback?

krassowski: I do not have a strong opinion on what we should use. It just feels too complex and hard to reason about if the fallback has fallbacks - if I am reading it correctly.

tests/test_core_path.py
Darshan808: It's testing that when a legacy `@jupyterlab/builder` dependency is present and no core version is explicitly provided, we use the builder's version instead of resolving `"latest"`.

Yes `@jupyterlab/core-meta@4.5.7` isn't on npm but it's monkeypatched here, so we're just verifying that the code *attempts* to fetch that version rather than `"latest"`.

In the real code path, that npm request would return a 404, which would then trigger the GitHub fallback and fetch the corresponding `core-meta` from GitHub instead.

krassowski (approved):
Let's merge to release now and iterate in follow-up

emilykl:
FYI I think this change is causing a build failure for us in Plotly when the `@jupyterlab/builder` version spec includes `||`, e.g.

```
"@jupyterlab/builder": "^4.3.6 || ^3.6.8"
```

<details>

<summary>Click to show full stack trace from log file</summary>

```
[33m@jupyterlab/core-meta was not found in node_modules. This extension declares a devDependency on @jupyterlab/builder@4.3.6 || ^3.6.8, which is a legacy package so core-meta 4.3.6 || ^3.6.8 will be used instead of the latest release.  To avoid this, add @jupyter/builder as a devDependency instead of @jupyterlab/builder.
 [0m
Traceback (most recent call last):

  File "/Users/ekl/code/plotly.py/venv/_fresh-2026-08-07/lib/python3.14/site-packages/jupyter_builder/debug_log_file_mixin.py", line 52, in debug_logging
    yield

  File "/Users/ekl/code/plotly.py/venv/_fresh-2026-08-07/lib/python3.14/site-packages/jupyter_builder/base_extension_app.py", line 53, in start
    self.run_task()
    ~~~~~~~~~~~~~^^

  File "/Users/ekl/code/plotly.py/venv/_fresh-2026-08-07/lib/python3.14/site-packages/jupyter_builder/extension_commands/build.py", line 63, in run_task
    build_labextension(
    ~~~~~~~~~~~~~~~~~~^
        self.extra_args[0],
        ^^^^^^^^^^^^^^^^^^^
    ...<6 lines>...
        core_path=self.core_path or None,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^

  File "/Users/ekl/code/plotly.py/venv/_fresh-2026-08-07/lib/python3.14/site-packages/jupyter_builder/federated_extensions.py", line 232, in build_labextension
    core_package_file = core_package_file or get_core_meta(
                                             ~~~~~~~~~~~~~^
        core_version,
        ^^^^^^^^^^^^^
        ext_path=ext_path,
        ^^^^^^^^^^^^^^^^^^
        logger=logger,
        ^^^^^^^^^^^^^^
    )
    ^

  File "/Users/ekl/code/plotly.py/venv/_fresh-2026-08-07/lib/python3.14/site-packages/jupyter_builder/core_path.py", line 90, in get_core_meta
    _download_npm_core_meta(npm_version, npm_cache_file)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/Users/ekl/code/plotly.py/venv/_fresh-2026-08-07/lib/python3.14/site-packages/jupyter_builder/core_path.py", line 393, in _download_npm_core_meta
    metadata = json.loads(_http_get(f"{JPBLD_NPM_URL}/@jupyterlab/core-meta/{version}"))
                          ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/Users/ekl/code/plotly.py/venv/_fresh-2026-08-07/lib/python3.14/site-packages/jupyter_builder/core_path.py", line 55, in _http_get
    with urllib.request.urlopen(req, timeout=timeout) as resp:  # noqa: S310
         ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/urllib/request.py", line 187, in urlopen
    return opener.open(url, data, timeout)
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^

  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/urllib/request.py", line 487, in open
    response = self._open(req, data)

  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/urllib/request.py", line 504, in _open
    result = self._call_chain(self.handle_open, protocol, protocol +
                              '_open', req)

  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/urllib/request.py", line 464, in _call_chain
    result = func(*args)

  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/urllib/request.py", line 1369, in https_open
    return self.do_open(http.client.HTTPSConnection, req,
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                        context=self._context)
                        ^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/urllib/request.py", line 1321, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
              encode_chunked=req.has_header('Transfer-encoding'))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/http/client.py", line 1367, in request
    self._send_request(method, url, body, headers, encode_chunked)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/http/client.py", line 1378, in _send_request
    self.putrequest(method, url, **skips)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/http/client.py", line 1212, in putrequest
    self._validate_path(url)
    ~~~~~~~~~~~~~~~~~~~^^^^^

  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/http/client.py", line 1312, in _validate_path
    raise InvalidURL(f"URL can't contain control characters. {url!r} "
                     f"(found at least {match.group()!r})")

http.client.InvalidURL: URL can't contain control characters. '/@jupyterlab/core-meta/4.3.6 || ^3.6.8' (found at least ' ')
```

</details>


Ultimately I'll just take this as a hint to switch to `@jupyter/builder` but wanted to give you a heads-up.

(See https://github.com/plotly/plotly.py/pull/5694)

PR #155, opened 2026-07-29
Switch back to Module Federation v1

### LINKED ISSUE #154: jupyter-builder unintentionally switched to Module Federation v2 with stricter range semantics
## Description

The issue discussed in:
- https://github.com/jupyterlab/jupyter-chat/issues/471
- https://github.com/jupyterlab/jupyter-chat/pull/472#issuecomment-5123140912
- #152

comes from #54 pulling Module Federation v2. This was not an intentional choice, just something that slipped.

I think we should revert that and release a new version ASAP

krassowski:
Fixes #154


src/extensionConfig.ts
```diff
@@ -11,7 +11,18 @@ import * as glob from 'glob';
...
+// `rspack.container.ModuleFederationPlugin`, which uses the Module Federation
+// 2.0 runtime. MF2 resolves a shared package that is consumed with
+// `import: false` and `singleton: false` - which is how core packages absent
+// from JupyterLab's `singletonPackages` are consumed, e.g.
+// `@jupyterlab/docregistry` - by failing hard when no version in the share
+// scope satisfies `requiredVersion`, since there is no bundled fallback to fall
+// back to. V1 keeps webpack's behaviour of warning and using whatever version
+// the host provides, which is what makes an extension built against one
+// JupyterLab minor loadable in the next.
+const { ModuleFederationPluginV1: ModuleFederationPlugin } = rspack.container;
```
dlqqq, line 25: Hmm, is there a way to do this without going back to MF1? Have a gut feeling that this is not the right thing to do long-term -- you all have done a lot in `jupyter-builder` to modernize the build experience and this seems like a step back. MF v1 seems like it is no longer maintained.

I don't know much about MF but I believe there's some `strictVersion` configuration we can add that says "hey, try to load this anyways even if the version ranges don't work out". I think it might be better to try that first and avoid rushing this in.

But also please don't block this on my feedback! You all are the experts. I'll need a bit more time to get caught up on MF before I can give more detailed feedback.

dlqqq (commented):
Thank you for putting this together so quickly! Do whatever you think is best, I need more time to get accustomed to MF and also play around with `jupyter-builder` more. Really cool that you all have migrated to `rspack`.

jtpio (approved):
Thanks!

src/extensionConfig.ts
krassowski: It is not a step back; we never intentionally migrated to MF2. Changes like that should be discussed and agreed upon, if we moved from MF1 to MF2 by accident that needs to be reverted ASAP as it breaks the pre-existing contract of extension compatibility.

`strictVersion` does not help here:
1. Problematic packages (`docregistry`, `nbformat`, `observables`) already end up with `strictVersion: false` as they are non-singletons without fallbacks (with `import: false`). Most other packages have `strictVersion: false` but either have fallbacks (e.g. `@jupyter-notebook/application`) or are singletons (the rest).
2. Switch to MF2 made `strictVersion: true` the behaviour regardless of the setting for all non-singletons; this looks like an unaddressed bug in MF2 https://github.com/module-federation/core/issues/4651
3. `strictVersion` is missing from MF2's documented `SharedConfig` (https://module-federation.io/integrations/bundler/metro#sharedconfig-interface); the runtime only honours it for singletons and ignores it for everything else; you may have been looking at https://webpack.js.org/plugins/module-federation-plugin/ (webpack/MF1)

We can move back to MF2 once this is addressed.

PR #157, opened 2026-07-31
Make module federation version configurable

Darshan808:
## Make the Module Federation version configurable

MF1 was hardcoded in `extensionConfig.ts` by #155. This makes the MF1/MF2 choice an option, **defaulting to MF1**.

### New flag

| Layer | Option |
| --- | --- |
| CLI (`build`, `watch`) | `--module-federation-version=<1\|2>` |
| Extension `package.json` | `jupyterlab.moduleFederationVersion: 1 \| 2` |
| Python API | `build_labextension(..., module_federation_version=None)` (same for `watch_labextension`) |
| Node builder | `generateConfig({ moduleFederationVersion })` |

Invalid values are rejected with a clear error; the option is refused outright for the legacy `@jupyterlab/builder` marker, which has no such flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,13 +57,14 @@ jlpm build
 <details>
 <summary><code>build</code></summary>
 
-| Flag                         | Description                                                            |
-| ---------------------------- | ---------------------------------------------------------------------- |
-| `--development`              | Build in development mode (default: `False`)                           |
-| `--source-map`               | Generate source maps (default: `False`)                                |
-| `--static-url=<url>`         | Set the URL for static assets                                          |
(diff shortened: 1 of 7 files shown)


Darshan808:
I'm not sure how to test it properly. When I try to reproduce the original issue, it's only visible in the browser's console. Any thoughts on what kind of test could be suitable here ?

Darshan808:
I think we will able to safely switch to mf v2, if we decide to move forward with https://github.com/jupyterlab/jupyterlab/pull/19251 in lab.

PR #159, opened 2026-07-31
Support partial version specifiers (`4`, `4.5`)

### LINKED ISSUE #158: Partial version specifiers like `4` or `4.5` fail to resolve `core-meta`
### Description

`4` and `4.5` are valid npm version specifiers, but `jupyter-builder` treats them as concrete versions and builds URLs that can't exist.

With `"@jupyterlab/builder": "^4.5"` (for example, [referenced here](https://github.com/geojupyter/jupytergis/issues/1700)), it requests:

- npm: `https://registry.npmjs.org/@jupyterlab/core-meta/4.5`
- GitHub: `https://raw.githubusercontent.com/jupyterlab/jupyterlab/v4.5/jupyterlab/staging/package.json`

Both endpoints require a full `M.N.P` version, so resolution fails regardless of whether npm or GitHub is used.

Darshan808:
## Description

Expand partial npm version specifiers to the wildcard form already supported by the existing range resolution logic, allowing them to resolve to the highest matching release instead of constructing invalid `M.N` URLs.

Specifically:

- `4.5` → `4.5.x`
- `4` → `4.x.x`
- `*` → `latest`

This applies to both:

- the explicit `--version` argument, and
- legacy `@jupyterlab/builder` pins such as `^4.5`.

### DIFF
--- a/jupyter_builder/core_path.py
+++ b/jupyter_builder/core_path.py
@@ -34,6 +34,10 @@ _MAX_GITHUB_TAG_PAGES = 10
 #: Pre-migration marker package.
 _LEGACY_BUILDER_MARKER = "@jupyterlab/builder"
 
+#: A version specifier built only from numeric and wildcard components, e.g. "4",
+#: "4.5", "4.x" or "4.5.*". These are npm ranges rather than concrete versions.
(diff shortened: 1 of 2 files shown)


krassowski (approved):
Thank you!

PR #164, opened 2026-08-06
Fix path issues when building on Windows

### LINKED ISSUE #163: Windows: glob 13 causes remoteEntry lookup to fail and writes _build.load: "static"
## Description

On Windows, with the current copier template, building and running an extension fails at runtime because jlab is unable to load the javascript:

`GET http://localhost:8889/lab/extensions/<extension_name>/static NS_ERROR_CORRUPTED_CONTENT`

It seems like this is due to 
```javascript
const files = glob.sync(
  path.join(staticPath, 'remoteEntry.*.js')
);
```
which, with version > 7 of glob results in C:\...\static\remoteEntry.*.js, but the backslashes are considered escape characters, so the route becomes just '/static/'

## Reproduce

On Windows:

1. `copier copy --trust https://github.com/jupyterlab/extension-template .`
2. `jlpm install`
3. `pip install -e .`
4. `jupyter lab` (check console for error)

Forcing the resolution in package.json of glob back to version 7 allows the extension to load without issue.
```
"@jupyter/builder@npm:1.2.1/glob": "npm:7.2.3",
```

## Expected behavior

Extension should resolve path `static/remoteEntry.<hash>.js` and load without error.

## Context

- Operating System and version: Windows 11
- Browser and version: Firefox 152.0.6
- JupyterLab version: 4.6.2

krassowski:
- [x] Adds a test for #163 (should fail on CI)
- [x] Fixes #163
- [x] Fixes symlink creation on Windows for editable watch build


Darshan808 (approved):
Thank you

PR #166, opened 2026-08-09
Support npm range specifiers in the `@jupyterlab/builder` version pin

### LINKED ISSUE #165: Support npm range specifiers in the `@jupyterlab/builder` version pin
> FYI I think this change is causing a build failure for us in Plotly when the `@jupyterlab/builder` version spec includes `||`, e.g.
> 
> ```
> "@jupyterlab/builder": "^4.3.6 || ^3.6.8"
> ```

_Originally posted by @emilykl in https://github.com/jupyterlab/jupyter-builder/issues/153#issuecomment-5220864925_

Darshan808:
### Description

Range unions and compound ranges now reduce to a single version — highest alternative across `||`, lower bound within one: `^4.3.6 || ^3.6.8` → `4.3.6`, `>=4.3.6 <5.0.0` → `4.3.6`, `4.1.0 - 4.5.0` → `4.1.0`. Ordering is numeric and prefers a stable release over its own prerelease.


krassowski (approved):
Looks good, thank you!

To be more compatible with what the old builder would produce we could resolve:
- `~4.5.7` → `4.5.7`
- `^4.5.7` → `4.5.x`

but that only works for GitHub fallback. What do you think?

krassowski:
> ^4.5.7 → 4.5.x

The other downside is that the `>=4.5.7` restriction gets lost.

Darshan808:
> > ^4.5.7 → 4.5.x
>
> The other downside is that the >=4.5.7 restriction gets lost.

We can do this. Since wildcards are resolved to their highest matching version, I believe the `>=` will still be preserved.
https://github.com/jupyterlab/jupyter-builder/blob/c4c7fcdca77bdc8d08ab573d1bdf0533d4b2e91a/jupyter_builder/core_path.py#L343-L344

krassowski:
Let's do that then :)

jupyter_builder/core_path.py
```diff
@@ -169,15 +171,61 @@ def _legacy_builder_marker_version(ext_path: Path) -> str | None:
...
+    name, or "workspace:*".
+    """
+    candidates = [
+        bounds for alternative in spec.split("||") if (bounds := _alternative_bounds(alternative))
+    ]
+    if not candidates:
+        return None
+    # Alternatives are ranked by the version each one starts at, but it is the
+    # requestable form of the winning alternative that gets resolved.
+    _, requested_version = max(candidates, key=lambda bounds: _semver_key(bounds[0]))
```
krassowski, line 201: Can you add a test ensuring against correct transitions between alpha → beta → rc? Opus says this will fail as of now

Darshan808: Ah, yes it is actually failing due to how versions are being sorted. Fixed in following commit.

PR #167, opened 2026-08-14
Add a `jupyterlab.rspackConfig` alias

agriyakhetarpal:
Hi! I came across this while migrating to `jupyter-builder`. I find `jupyterlab.webpackConfig` to be a bit misleading, since the build now runs with Rspack. This is a small PR that adds an `rspackConfig` field to the metadata schema as an alias for `webpackConfig`; it falls back to `webpackConfig` if not set and takes precedence if both are set.

Would this change be desirable? I am happy to add some tests for this, if so.

xref agriyakhetarpal/jupyterlite-pdf-exporter#46


krassowski:
Hmm this looks like this way of keeping `webpackConfig` was decided in https://github.com/jupyterlab/jupyterlab/pull/16005 (I don't see a discussion but the [diff](https://github.com/jupyterlab/jupyterlab/pull/16005/changes#diff-e38d93b95b69356f9cb7b25f09ffe2c227ee359cf4a11a6d4f9aa8c4865261b1) modified https://jupyterlab.readthedocs.io/en/latest/extension/extension_dev.html#custom-rspack-config).

So, three thoughts:
- I find it more intuitive to rename it to rspack and keep the old one as compatibility fallback
- this would add a new public API and we will need to support it realistically forever
- if we want to make this change, we should update the docs

agriyakhetarpal:
Awesome, thanks, @krassowski! I think it would be fine to support this key forever, unless y'all aren't happy with Rspack and are considering switching to another bundler sometime soon. I've added a couple of tests and can follow up with a docs PR to JupyterLab when this is released (or do you want one now?).

Darshan808 (approved):
Looks good to me. Thanks @agriyakhetarpal. We can follow up with a docs PR to JupyterLab later when this is released.

PR #168, opened 2026-08-15
Remove licence classifier (PEP 639)

mwtoews:
License classifiers are deprecated since [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-classifiers), so remove it. The existing `License-Expression` does a better job of describing this project's multi-licencing.


krassowski (approved):
Thank you!

PR #175, opened 2026-09-04
Generate license report for the bundled `yarn.js` and correct license in `pyproject.toml`

### LINKED ISSUE #169: Licenses and versions of bundled dependencies
In order to properly package this in Fedora Linux, I need to know the licenses and versions of bundled dependencies. There are a couple of issues I see now:

- There are packages mentioned in yarn.js (like typanion, yup, clipanion, …) but not included in the file with third-party licenses.
- The license of Yarn itself is not there.
- There is no info about versions of the bundled packages.

Would it be possible to add a generated file like `third-party-licenses.json` in Jupyterlab here?

I don't know the build process for JS libraries, so I'm sorry if I'm missing something.

Darshan808:
### Description

`jupyter_builder/yarn.js` bundles 189 npm packages, but we only documented 6 of them and the license file pointed at the wrong repo (`yarnpkg/yarn` instead of `yarnpkg/berry`) with no versions. Same deal with the vendored semver: stale URL, no version.

Adds `scripts/generate_yarn_licenses.py`, which checks out berry at the tag matching the bundle, installs the prod deps, and walks the tree. It writes:

- `THIRD_PARTY_LICENSES/yarn.js.third-party-licenses.json` — same schema `JSONLicenseWebpackPlugin` already emits, so it looks like JupyterLab's `third-party-licenses.json`. Includes versions, which was requested in #169 for `Provides: bundled(npm(...))` and CVE tracking.

- `THIRD_PARTY_LICENSES/yarn.js.LICENSE.txt` — now rendered from the JSON instead of hand-maintained.
- the `license` expression in `pyproject.toml`, recomputed from every licenseId found

That last one turned up four licenses we were missing:

before: `BSD-2-Clause AND BSD-3-Clause AND ISC AND MIT`
after:  `0BSD AND Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND ISC AND MIT AND Unlicense AND (MIT OR GPL-2.0-only)`

CI re-runs the generator and fails on a diff. Re-vendoring steps are in RELEASE.md.


pyproject.toml
```diff
@@ -9,8 +9,8 @@ build-backend = "hatchling.build"
 name = "jupyter_builder"
 description = "JupyterLab build tools"
 readme = "README.md"
-license = "BSD-3-Clause AND MIT AND ISC"
-license-files = ["LICENSE", "THIRD_PARTY_LICENSES/*.txt"]
+license = "0BSD AND Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND ISC AND MIT AND Unlicense AND (MIT OR GPL-2.0-only)"
```
krassowski: This lists three licenses from packages that do not contribute to the bundle:
- 0BSD comes only from `tslib` 1.13.0,
- Unlicense only from `stream-buffers` 3.0.2,
- (MIT OR GPL-2.0-only) only from `node.extend` 2.0.2.

`esbuild` inlines its own TypeScript helpers so `tslib` is never imported, and the other two are not reached from any bundled entry point.

THIRD_PARTY_LICENSES/yarn.js.third-party-licenses.json
krassowski, line 1: Two packages whose code is in the shipped jupyter_builder/yarn.js are absent from the report: `arg` 5.0.2 and `resolve.exports` 1.1.0, both MIT.

krassowski (commented):
It looks like it misses a few things on one hand and is over-eager on the other one. I will see if we can do it better.

krassowski (approved):
I think that's good enough, thanks!

PR #176, opened 2026-09-04
Allow Node.js prereleases in the version check

### LINKED ISSUE #174: The node version message is misleading when newer Node.js is installed
## Description

Just spotted

```
Building this extension requires Node.js ^20.19.0 || >=22.12.0 (found v26.8.0-alpha.0.0.0). Please upgrade Node.js.
```

https://github.com/jupyterlab/jupyterlab/actions/runs/33617281759/job/100205938621?pr=19581#step:8:8986

Darshan808:
### Description
The Node.js version check now accepts prereleases. npm-style semver never matches a prerelease against a range written for stable releases, so any prerelease failed the check no matter how new — Node 26.8.0 shipped reporting itself as `v26.8.0-alpha.0.0.0` (nodejs/node#65566), and we errored saying "upgrade" from it.

Also reworded the message to "Please install a compatible Node.js version.", since "upgrade" was wrong here.

### DIFF
--- a/jupyter_builder/federated_extensions.py
+++ b/jupyter_builder/federated_extensions.py
@@ -359,14 +359,24 @@ def _read_rspack_node_range(builder: str, ext_path: str) -> str:
 
 
 def _satisfies_allowing_prerelease(current: str, node_range: str) -> bool:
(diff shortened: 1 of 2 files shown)


krassowski (approved):
Thanks!

PR #178, opened 2026-09-07
Fix flaky macOS CI tests by authenticating GitHub API calls

### LINKED ISSUE #177: Tests on macos are flaky on CI
See for example https://github.com/jupyterlab/jupyter-builder/actions/runs/33907262119/job/101135029703

Yash990-bit:
I looked into the flaky CI failures in test_watch_functionality_jupyterlab_builder and test_files_build_jupyterlab_builder, mainly seen on macOS runners.

The issue was caused by jupyter-builder making unauthenticated requests to the GitHub API when resolving the @jupyterlab/core-meta version. On shared GitHub Actions runners, these requests can hit GitHub's rate limit and return a 403: rate limit exceeded error, which makes the tests fail.

To fix this, I updated _http_get() to use the GITHUB_TOKEN when making requests to the GitHub API. I also exposed GITHUB_TOKEN in the builder and downstream test workflows.

This should prevent the rate-limit failures and make the CI tests more stable across different runners.

### DIFF
--- a/.github/workflows/builder-tests.yml
+++ b/.github/workflows/builder-tests.yml
@@ -13,6 +13,8 @@ defaults:
   run:
     shell: bash -eux {0}
 
+permissions:
+  contents: read
 
 jobs:
   build:
--- a/.github/workflows/builder-tests.yml
+++ b/.github/workflows/builder-tests.yml
@@ -13,8 +13,6 @@ defaults:
   run:
     shell: bash -eux {0}
 
-env:
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
   build:
@@ -96,6 +94,8 @@ jobs:
         run: pip install -e ".[test,dev]"
       - name: Run the tests
         run: pytest --cov
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - uses: jupyterlab/maintainer-tools/.github/actions/upload-coverage@v1
 
   test_lint:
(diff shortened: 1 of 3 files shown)


.github/workflows/builder-tests.yml
```diff
@@ -13,6 +13,9 @@ defaults:
   run:
     shell: bash -eux {0}
 
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
krassowski: There is no reason to pass it on the top-level

Yash990-bit: Thanks for pointing this out. You're right — there’s no need to expose `GITHUB_TOKEN` at the workflow level when only the test step needs it. I’ve moved it to the `Run the tests` step so the token is only available where it’s actually required.

.github/workflows/downstream-tests.yml
```diff
@@ -3,6 +3,9 @@ name: Downstream Tests
 on:
   pull_request:
 
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
krassowski: Why is it needed here?

Yash990-bit: Yes — it's not needed here at all. Removed it entirely from downstream-tests.yml

krassowski (approved):
Thank you!

Yash990-bit:
"Glad I could help! Thanks for the quick review and merge.

PR #179, opened 2026-09-11
Record plugin IDs as static metadata

krassowski:
This PR exposes plugin IDs as static metadata which enables accessing the metadata of a plugin before the extension gets loaded.

https://github.com/jupyterlab/jupyterlab/pull/19627 deferred loading (fetching via network) extensions that are disabled. It could not:
- a) defer loading an extension that was disabled by disabling all of its plugins (rather than disabled by disabling it by name)
- b) skip loading them altogether because the information about the extension needs to be shown in the Plugin Manager

This PR will enable (a).

For an extension `@foo/ext` with plugins `@foo/ext:a` and `@foo/ext:b`:
- for `disabledExtension: ['@foo/ext']` https://github.com/jupyterlab/jupyterlab/pull/19627 worked well
- for equivalent `disabledExtension: ['@foo/ext:a', '@foo/ext:b']` it did not actually defer

It also opens the door to (b) - we would need to add `description` and `autoStart` both are extractable statically. Maybe in a follow-up PR once this gets decided; this is why using the list of objects rather than list of strings here.


krassowski:
This has two limitations:
1. Extracting plugin data from plugins that relay on browser-only symbols is not reliable if using evaluation. For example, extraction from `@jupyterlab/help-extension` aborts because it attempts to read on module level from [`window.location.protocol`](https://github.com/jupyterlab/jupyterlab/blob/c9277f586523f97ec65756bcdc2ade53adbf25e6/packages/help-extension/src/index.tsx#L59) and `window` is `undefined` in Node.js context so the accessor throws
   > this is a recently introduced buglet (when we were fixing a vulnerability) - JupyterLab core _should_ import in both browser and Node.js runtime per established practice; if that was in a non `-extension` package I would even call the `window.location.protocol` access in that place a bug. However, extensions are free to expect browser-only runtime, and we need to handle both.
2. If plugins are added dynamically (conditionally) to the exported list, the metadata list will include all possible plugins declared, regardless of the conditions. For example, the `sometimes` plugin will be always extracted:
    ```ts
    import { PageConfig } from '@jupyterlab/coreutils';
    
    const always = {
      id: 'bad:always',
      description: 'Always provided.',
      autoStart: true,
      activate: () => {
        console.warn('always activated');
      }
    };
    
    const sometimes = {
      id: 'bad:sometimes',
      description: 'Provided unless the deployment opts out.',
      autoStart: true,
      activate: () => {
        console.warn('sometimes activated');
      }
    };
    
    const plugins = [always];
    
    // The deployment can opt out of the second plugin. Nothing sets this option, so
    // in a browser `getOption` returns '' and the plugin is provided.
    if (!PageConfig.getOption('demoOptOut')) {
      plugins.push(sometimes);
    }
    
    export default plugins;
    ```

krassowski:
Switched to static analysis for extraction. This is 10x faster, safer, side-effect free and does not have the problem with extensions that expect browser-only symbols. The only downside is adding a runtime dependency on `acorn` (alternatively we could use `typescript` or `@swc/core`) - which was already a transitive dep.

